PR #1, opened 2017-10-01
Small improvements

laurensmartina:


system/datatypes/text.php
```diff
@@ -20,6 +20,7 @@ function __construct($pageListNode, $args) {
 			registerCommandCallback('textRevert', Array($this, 'revert'));
 		}
 
+		/** @deprecated  */
```
matthijskooijman: This should probably not go into this commit, the commit message does not mention this change. Also, why is this deprecated?

system/datatypes/festival.php
```diff
@@ -231,7 +231,7 @@ function showContributions() {
 		function getSettingsForm() {
 $html = <<<'EOF'
 <table>
-	<tr><td><label for="festival-title">Festival title</label> *</td><td><input name="festival-title"/></td></tr>
+	<tr><td><label for="festival-title">Festival title</label> *</td><td><input id="festival-title" name="festival-title"/></td></tr>
```
matthijskooijman: Nitpick: This is a bit of a different change from the rest of the typo fixes, so this might be better placed into its own commit? Not essential, though.

system/datatypes/text.php
laurensmartina: You are right, this should go out, I'll fix it.

system/datatypes/festival.php
laurensmartina: True, you are right. But let's not do it now, I'll commit more carefully next time.

matthijskooijman:
Looks good! I'd like to see that last commit merged into the first one to clean up the history. You'll need interactive rebase for that. I was going to show you how that works, so this is probably a good example. Let's look at this next tuesday at our meeting?

laurensmartina:
Perfect, let's do that!

matthijskooijman:
Looks perfect now, merging.

PR #2, opened 2017-10-01
Mailinglist

laurensmartina:


system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+			if (empty($form->errors)) {
+				// check if email is already in the list, no need to have it twice.
+				$email = strtolower($form->dataFor(self::FIELD_NAME_EMAIL));
+				$addresses = $this->getDoc()->getOrCreate('addresses');
+				$address = $addresses->children()->findXPath('//address[@email="' . $email . '"]')->first();
+
+				if (!$address instanceof HyphaDomElement) {
+					$addAddress = true;
+					$sendEmail = true;
+					$code = uniqid();
```
matthijskooijman: I believe `uniqid()` is based on the current timestamp and thus is fairly predictable. I the festival module, I used `bin2hex(openssl_random_pseudo_bytes(8))` to generate a more proper random identifier instead, which might also be useful here.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+				} else {
+					// if the email is there just send another email if it is still pending
+					$sendEmail = $address->getAttribute('status') == self::ADDRESS_STATUS_PENDING;
+					$addAddress = false;
+					$code = $address->getAttribute('code');
+				}
+
+				if ($addAddress) {
+					// add email with status pending
+					$this->xml->lockAndReload();
```
matthijskooijman: After taking the lock, this should check again if the address isn't already present (and skip adding the address if so). Alternatively, the lock can be moved up to include the checks there.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+					$address->setAttribute('email', $email);
+					$address->setAttribute('status', self::ADDRESS_STATUS_PENDING);
+					$address->setAttribute('confirm-code', $code);
+					$this->getDoc()->getOrCreate('addresses')->append($address);
+					$this->xml->saveAndUnlock();
+				}
+
+				if ($sendEmail) {
+					// send email so that action can be confirmed
+					$link = $this->generateFullPath(sprintf('%s/confirm?email=%s&code=%s', $this->pagename, $email, $code));
```
matthijskooijman: I think that urlencoding the variables interpolated into the url would be appropriate here? Possibly the e-mail validation regex already rules out any problematic characters, but encoding things here would certainly rule out problems.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+					$subject = hypha_getTitle() . ' - ' . __('ml_confirm_email_subject');
+					$this->send($subject, '<a href="' . $link . '">Please confirm your email address</a>', [$email]);
+
+					notify('success', __('ml_successfully_submitted_email'));
+				} else {
+					notify('success', __('ml_successfully_subscribed'));
```
matthijskooijman: I wonder if here (and some other place below) it might be better to add an "already subscribed" (or below, "already confirmed") message, to give the user a bit more feedback about what is (not) happening.

system/core/WymHTMLForm.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+require_once __DIR__ . '/document.php';
+
+class WymHTMLForm extends HTMLForm
```
matthijskooijman: This commits looks great!

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+						return $this->showMailingAction($firstArgument);
+				}
+		}
+	}
+
+	private function loadDictionary() {
+		// TODO [LRM]: move dictionary to language files
+		global $hyphaUser, $hyphaDictionary;
+
+		$language = $hyphaUser ? $hyphaUser->getAttribute('language') : hypha_getDefaultLanguage();
```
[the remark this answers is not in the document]
laurensmartina: Nice catch, this code was from before I dug into the structure.. I'll fix it.

system/datatypes/mailinglist.php
laurensmartina: In this case I don't really mind that the code is not unique or is predictable because I use it in combination with the email address, that being said, I have no objections to use your suggested method.
Note: `uniquid()` is currently used in `HyphaDomElement::generateId()` perhaps this should be updated as well, `uniquid` is not 100% unique which could cause problems.

system/datatypes/mailinglist.php
laurensmartina: Good point, I'm so used to frameworks that cover everything that I'm forgetting these kinds of things.

system/datatypes/mailinglist.php
laurensmartina: The current behaviour is based on a design principle (I forgot the name) that the output should always be consistent. Also the receiving party does not care that it is 'already subscribed', it just cares that it got 'subscribed' (of confirmed or unsubscribed). We can talk about it if you want.

system/datatypes/mailinglist.php
laurensmartina: I've moved the lock up in the code, otherwise all the same checks needed to be done again.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+		if (!$email || !$code) {
+			return notify('error', __('missing-arguments'));
+		}
+
+		$address = $this->getDoc()->findXPath('//address[@email="' . $email . '"][@unsubscribe-code="' . $code . '"]')->first();
+		if (!$address instanceof HyphaDomElement) {
+			return notify('error', __('invalid-email-code-combination'));
+		}
+
+		if ($address->getAttribute('status') != self::ADDRESS_STATUS_CONFIRMED) {
```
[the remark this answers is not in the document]
laurensmartina: moved the lock before fetching the address, in both cases.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+			return notify('error', __('email-has-invalid-status'));
+			$this->redirect('');
+		}
+
+		$id = $address->getId();
+		$this->xml->lockAndReload();
+		$address = $this->xml->document()->getElementById($id);
+		$this->getDoc()->getOrCreate('addresses')->removeChild($address);
```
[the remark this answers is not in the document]
laurensmartina: Although I'm not a big fan of communicating id's, I do think this is a good solution. But I would rather communicate 'unsubscribed' instead of 'already unsubscribed'.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+		if ($inDraft) {
+			// buttons
+			$commands = $this->findBySelector('#pageCommands');
+			$this->addSpacer($commands);
+			$commands->append($this->createButton(__('edit'), $mailingId . '/edit'));
+		}
+	}
+
+	private function createSubscribeForm(array $data = []) {
+		$subject = __('ml_email');
```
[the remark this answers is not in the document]
laurensmartina: nice catch! :-)

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+		$html = <<<EOF
+			<div class="section" style="padding:5px; margin-bottom:5px; position:relative;">
+				<label for="$emailFieldName">$subject</label>: <input type="text" id="$emailFieldName" name="$emailFieldName" placeholder="email address" />
+			</div>
+EOF;
+		/** @var \DOMWrap\Element $elem */
+		$elem = $this->html->createElement('form')->html($html);
+
+		// buttons
+		$elem->find('#' . $emailFieldName)->append($this->createButton(__('ml_subscribe'), null, self::FORM_CMD_LIST_SUBSCRIBE));
```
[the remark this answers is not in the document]
laurensmartina: hahaha, you are right. It worked, but is not really nice. I fixed it.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+	}
+
+	private function isPosted($cmd) {
+		return 'POST' == $_SERVER['REQUEST_METHOD'] && $cmd == $_POST['command'];
+	}
+
+	private function addSpacer(DOMWrap\NodeList $nodeList) {
+		$nodeList->append('<span style="padding: 0 5px;">|</span>');
+	}
+
```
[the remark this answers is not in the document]
laurensmartina: true, I'm just not sure where to place it.. I still need to get a better sense of the landscape.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+		$headers .= 'Content-type: text/html; charset=utf-8' . PHP_EOL;
+		$headers .= 'From: ' . $from . ' <' . $from . '>' . PHP_EOL;
+		$headers .= 'Reply-To: ' . $from . PHP_EOL;
+		$headers .= 'X-Mailer: PHP/' . phpversion();
+
+		foreach ($receivers as $receiver) {
+			mail($receiver, $subject, $messageHtml, $headers . PHP_EOL . 'To: ' . $receiver . PHP_EOL);
+			usleep(rand(500000, 2000000));
+		}
+	}
```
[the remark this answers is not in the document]
laurensmartina: If `sendMail()` is going to be used, I think it needs to be unified with the method used in this case.
I'll replace `PHP_EOL` with `\r\n`.

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+
+		// throw error if status is draft without client logged in
+		if ($inDraft && !isUser()) return notify('error', __('not-found')); // 404
+
+		// display mailing
+		$date = $mailing->getAttribute('date');
+		$main = $this->findBySelector('#main');
+		$main->append('<div><h2>' . $mailing->getAttribute('subject') . '</h2></div>');
+		if ($date) {
+			$main->append('<div>date: ' . (new \DateTime($date))->format('d-m-Y') . '</h2></div>');
```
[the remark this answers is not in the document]
laurensmartina: Formatting date without the intl extension is a pain.. We could however set the formatting in the translation files to simplify things (althoug it doesn't belong there). What do you think?

system/core/WymHTMLForm.php
laurensmartina: Thank you! :-)

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,728 @@
...
+					/** @var HyphaDomElement[] $addresses */
+					$addresses = $this->getDoc()->findXPath('//address[@status="' . self::ADDRESS_STATUS_CONFIRMED . '"]');
+					$receivers = [];
+					foreach ($addresses as $address) {
+						$receivers[] = $address->getAttribute('email');
+					}
+					$this->send($mailing->getAttribute('subject'), $mailing->getHtml(), $receivers, $mailing->getAttribute('sender'));
+
+					// mark mailing as sent
+					$this->xml->lockAndReload();
```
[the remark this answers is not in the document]
laurensmartina: implemented the sending/sent part.

we would need to think of a 'simple' solution in order to prevent a time out. simple chunking it will not de the trick imo. Something to discuss. I'll add a todo.

system/datatypes/mailinglist.php
matthijskooijman: > In this case I don't really mind that the code is not unique
I guess that (with significant brute-forcing) subscribing someone else to a mailinglist is not really a big security problem, but it won't hurt to make it a bit more secure.

> or is predictable because I use it in combination with the email address
That does not really help, since an attacker would just subscribe someone else's e-mail address and then know the address.

>, that being said, I have no objections to use your suggested method.
Yes, please.

> Note: uniquid() is currently used in HyphaDomElement::generateId() perhaps this should be updated as well, uniquid is not 100% unique which could cause problems.

Correct, but that's inside a loop that checks for uniqueness, so that should result in a unique id eventually. Using a random id could reduce chances of conflicts, but I think those are already slim enough (and now the ids are sorted in order of creation, which can be useful).

system/datatypes/mailinglist.php
matthijskooijman: Your argument would then be that the output is only based on the resulting state, not on the actions performed. It relates a bit to the concept of idempotency. I have no strong preference in this.

system/datatypes/mailinglist.php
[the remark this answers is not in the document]
matthijskooijman: I didn't mean to translate the date itself, but the `date: ` prefix.

system/datatypes/mailinglist.php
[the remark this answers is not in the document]
matthijskooijman: As before, I do not have a strong preference over 'unsubscribed' vs 'already unsubscribed'.

system/datatypes/mailinglist.php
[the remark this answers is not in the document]
matthijskooijman: Is there anything `sendMail` does not do that this method needs? This method of course generates the HTML content that is mailing-list-specific, so that part would stay in this method, but the generic parts could be handled by `sendMail()`?

system/datatypes/mailinglist.php
```diff
@@ -193,19 +198,20 @@ public function indexAction() {
...
 					$this->getDoc()->getOrCreate('addresses')->append($address);
 					$this->xml->saveAndUnlock();
+				} else {
+					$this->xml->unlock();
 				}
 
 				if ($sendEmail) {
 					// send email so that action can be confirmed
-					$link = $this->generateFullPath(sprintf('%s/confirm?email=%s&code=%s', $this->pagename, $email, $code));
+					$link = $this->generateFullPath(sprintf('%s/confirm?email=%s&code=%s', $this->pagename, urlencode($email), $code));
```
matthijskooijman: For consistency, I would urlencode all three parts. What if the pagename contains an `&`, or we switch to an code-generation method that outputs `%` or `&` or whatever? Just that something is not (directly) user-controlled, or previously verified / sanitized, is (IMHO) no reason to not encode it anyway.

system/datatypes/mailinglist.php
```diff
@@ -421,78 +421,80 @@ private function editMailingAction($mailingId = null) {
...
+		$saveMailing = $this->isPosted(self::FORM_CMD_MAILING_SAVE);
+		if ($this->isPosted() && !$sendMailing && !$saveMailing) {
+			return notify('error', __('ml_invalid-command'));
+		}
+		$isPosted = $sendMailing || $saveMailing;
+
 		// get mailing if id was given
 		/** @var HyphaDomElement $mailing */
-		if ($mailingId) {
+		if (null != $mailingId) {
```
matthijskooijman: Is this needed or better? Implict falseness test seems sufficient here, and I usually find it clear. A != null check might actually behave the same, due to the implicit conversions PHP does (perhaps you intended !== null?).

system/datatypes/mailinglist.php
```diff
@@ -421,78 +421,80 @@ private function editMailingAction($mailingId = null) {
...
 			$mailing = $this->xml->document()->getElementById($mailingId);
 			if (!$mailing instanceof HyphaDomElement) {
-				return notify('error', __('invalid-id-given'));
+				return notify('error', __('not-found')); // 404
 			}
 			// check if status is correct
-			if ($mailing->getAttribute('status') == self::MAILING_STATUS_SENT) {
+			$status = $mailing->getAttribute('status');
+			if (self::MAILING_STATUS_SENT == $status && $sendMailing) {
+				notify('success', __('ml_successfully_sent'));
```
matthijskooijman: This is another case where you either say "Mail was sent" or "Mail was already sent". Since sending a mail is actually an action, not just a state change, and you would rather not send duplicate mails, it seems more sensible to give more accurate feedback to the user. If a user somehow accidentally tries to send the mailing twice, they might think they sent the mailing list twice if the result is just "successfully sent". Alternately, if the sending fails (or at least the user thinks so) and they attempt to send it again (e.g. from a browser window they still had open from before it was originally sent), they might think the message was sent (again), when in reality nothing happened.

system/datatypes/mailinglist.php
```diff
@@ -151,6 +153,8 @@ private function loadDictionary() {
 				$hyphaDictionary['ml_if_unreadable_use_link'] = 'if unreadable, use link';
 				$hyphaDictionary['ml_no_sender'] = 'no sender, please edit the mailing list';
 				$hyphaDictionary['ml_description'] = 'description';
+				$hyphaDictionary['ml_missing_arguments'] = 'required arguments are missing';
+				$hyphaDictionary['ml_invalid_code'] = 'the code is invalide';
```
matthijskooijman: s/invalide/invalid/

system/datatypes/mailinglist.php
```diff
@@ -768,8 +762,40 @@ private function sendMailing($mailingId) {
...
+		return $senderData;
+	}
+
+	/**
+	 * @todo [LRM]: move so it can be used throughout Hypha
+	 *
+	 * @return string
+	 */
+	private function generateUniqueKey() {
+		return bin2hex(openssl_random_pseudo_bytes(8));
```
matthijskooijman: This does not necessarily make the key unique, randomness just makes it really unlikely. A loop (like in `generateId()`) could be added here, but that's tricky to write generically (e.g. what's the scope of the uniqueness?). So I'm not sure if I like this change, it might be better to just keep an id for each address?

matthijskooijman:
The extra changes look good (though I didn't look as closely as before). I added some more comments, perhaps we can talk about some of these tomorrow as well.

system/datatypes/mailinglist.php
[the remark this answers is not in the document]
laurensmartina: ... I'm a sloppy reader. :-)

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,959 @@
...
+					$addresses = $this->getDoc()->getOrCreate('addresses');
+					$addresses->append($address);
+					$this->xml->saveAndUnlock();
+				} else {
+					$this->xml->unlock();
+				}
+
+				if ($sendEmail) {
+					// send email so that action can be confirmed
+					$link = $this->constructFullPath(sprintf('%s/confirm?code=%s', $this->pagename, $code));
```
matthijskooijman: pagename and code need to be passed through `urlencode` here, since it's hard to guarantee they  contain no breaking special characters (e.g. &).

system/datatypes/mailinglist.php
```diff
@@ -0,0 +1,959 @@
...
+	 * @param null|string $command
+	 *
+	 * @return bool
+	 */
+	private function isPosted($command = null) {
+		return 'POST' == $_SERVER['REQUEST_METHOD'] && (null == $command || $command == $_POST['command']);
+	}
+
+	/**
+	 * @todo [LRM]: move so it can be used throughout Hypha
```
matthijskooijman: You should probably make this "and/or integrate this with `sendMail() and/or let it make use of `sendMail()`" (or perhaps just make it use `sendMail()` now already?)

matthijskooijman:
Looks good now, merged!

Issue #3, opened 2018-01-04
Mailinglist lacks compose view


giplt:
The mailinglist module doesn't seem to offer a way to compose and send a message.

PR #5, opened 2018-01-04
Bugfix issue #3

laurensmartina:
Fixes issue #3


matthijskooijman:
Code looks good, but:
 - There is an extra "merge PR #1" commit that seems to be duplicate in your history
 - The commit message is a bit brief. I usually prefer the first line to explain the change, since only referring to a bug report does not really tell you anything when looking through the commit history. The commit message should be self-contained and useful without referencing any external bug reports (which might not be available anymore in the future). If you refer to an issue, that should be extra. In this case, I would write "Closes #3." on a line by itself at the end of the message, to have [github automatically close the issue when it is merged]).

laurensmartina:
fixed

matthijskooijman:
Thanks. I'm still missing capitalization, and the message now no longer references the issue number at all?

PR #8, opened 2018-01-04
Small improvements to make codebase more robust

laurensmartina:
small improvements to make codebase more robust

check for output buffering when serving a file
check for variable before using
made ranking an anonymous function that is assigned to a variable
used currentRow property in HTMLTable

### DIFF
--- a/system/core/document.php
+++ b/system/core/document.php
@@ -502,6 +502,7 @@
 	}
 
 	class HTMLTable extends DOMWrap\Element {
+		/** @var HTMLTableRow */
 		private $currentRow;
 
 		function __construct() {
@@ -509,19 +510,19 @@
 		}
 
 		function addHeaderRow() {
-			$currentRow = new HTMLTableRow('th');
-			$this->appendChild($currentRow);
-			return $currentRow;
+			$this->currentRow = new HTMLTableRow('th');
+			$this->appendChild($this->currentRow);
+			return $this->currentRow;
 		}
 
 		function addRow() {
-			$currentRow = new HTMLTableRow('td');
-			$this->appendChild($currentRow);
-			return $currentRow;
+			$this->currentRow = new HTMLTableRow('td');
+			$this->appendChild($this->currentRow);
+			return $this->currentRow;
 		}
 
 		function addCell($contents = null) {
-			return $currentRow->addCell($contents);
+			return $this->currentRow->addCell($contents);
 		}
 	}
 
--- a/system/core/document.php
+++ b/system/core/document.php
@@ -181,8 +181,9 @@
 			foreach($postProcessingList as $func) $func($this);
 
 			// format <head> element into convenient order, could be removed to gain speed
-			function ranking($node) {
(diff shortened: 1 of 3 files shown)


system/core/document.php
```diff
@@ -499,26 +500,27 @@ function handleImageUpload($name, $fileinfo) {
...
-			$this->appendChild($currentRow);
-			return $currentRow;
+			$this->currentRow = new HTMLTableRow('td');
+			$this->appendChild($this->currentRow);
+			return $this->currentRow;
 		}
 
 		function addCell($contents = null) {
-			return $currentRow->addCell($contents);
+			return $this->currentRow->addCell($contents);
```
matthijskooijman: Did this code actually work before? Or is this really a bugfix?

laurensmartina: Looking at the code, it could not have worked properly; `$currentRow` was never defined.

laurensmartina:
> The output buffer must be started by `ob_start()` with `PHP_OUTPUT_HANDLER_CLEANABLE` and `PHP_OUTPUT_HANDLER_REMOVABLE` flags. Otherwise `ob_end_clean()` will not work.
- When serving a file, the `ob_start()` was never called before the `ob_end_clean()`.

> [http://php.net/manual/en/functions.anonymous.php](http://php.net/manual/en/functions.anonymous.php)
> [https://www.elated.com/articles/php-anonymous-functions](https://www.elated.com/articles/php-anonymous-functions)
- Assigning a anonymous function to a variable is pretty common and considered good practice in the php community. Defining a function like it was, sets the function in global scope. There is no need for the function to be in global scope and is a dangerous habit; if there was already a function with the same name this would cause a collision.

matthijskooijman:
> >   The output buffer must be started by ob_start() with PHP_OUTPUT_HANDLER_CLEANABLE and PHP_OUTPUT_HANDLER_REMOVABLE flags. Otherwise ob_end_clean() will not work.
 >   When serving a file, the ob_start() was never called before the ob_end_clean().

The docs seem a bit unclear about what happens exactly when you call `ob_end_clean()` without `ob_start()`, but it seems this will just cause `ob_end_clean()` fail and return false, which would be acceptable for what happens here.

Having said that, checking for an active output buffer using `ob_get_length()` might be a good practice. However, the current approach checks for a falsy return value of `ob_get_length()`. If output buffering is active, but no bytes were buffered yet, this returns 0 and output buffering is not terminated, which seems like a bug. When output buffering is not active, `ob_get_length()` returns false, so `if (ob_get_length() !== false)` seems better. Or, as I just learned from the docs that output buffering can be nested, even better might be:

    while (ob_get_level())
        ob_end_clean();

http://php.net/manual/en/function.ob-get-level.php

> Defining a function like it was, sets the function in global scope.

Oh, right, forgot that PHP doesn't really supported nested functions. Good change then (but still, this kind of info should probably go into the commit message).

laurensmartina:
fixed

PR #17, opened 2018-02-17
Added HyphaRequest object

laurensmartina:


system/core/HyphaRequest.php
```diff
@@ -0,0 +1,131 @@
...
+				$this->systemPage = $systemPage;
+			}
+		}
+
+		return $this->systemPage === false ? null : $this->systemPage;
+	}
+
+	public function getPageType() {
+		if ($this->isSystemPage() || $this->getLanguage() !== null) {
+			return null;
```
matthijskooijman: Doesn't an url with a language have a page type?

system/core/pages.php
```diff
@@ -94,139 +94,132 @@ function newPage() {
...
+		$language = $hyphaRequest->getLanguage();
+
 		// set wiki language. we want to store this in a session variable, so we don't loose language when an image or the settingspage are requested
-		if (count($args) > 0 && array_key_exists($args[0], $isoLangList)) {
-			$hyphaLanguage = $args[0];
-			array_shift($args);
-		} else {
-			$hyphaLanguage = hypha_getDefaultLanguage();
-		}
+        $hyphaLanguage = $language !== null ? $language : hypha_getDefaultLanguage();
```
matthijskooijman: It looks like there is an indentation error here.

system/core/HyphaRequest.php
```diff
@@ -0,0 +1,131 @@
...
+	public function getRequestQuery($excludeLanguage = true) {
+		return implode('/', $this->getRequestParts($excludeLanguage));
+	}
+
+	/**
+	 * @param bool $excludeLanguage
+	 *
+	 * @return array
+	 */
+	public function getRequestParts($excludeLanguage = true) {
```
matthijskooijman: Would it make sense to also remove the system pagename and page type from this list? Or perhaps add a new `getArgs()` method that returns all the "extra" request parts (e.g. everything that is not accessible through the other get* methods)?

This would allow system page loading below to handle `$args[0]`, now it skips the first element since that is the system page type (which is a bit of an abstraction leak).

index.php
```diff
@@ -94,15 +94,16 @@
 		Check is a user is logged in and load user data. Add login/logout functionality according to session login status.
 	*/
 
-	$request = explode('/', $hyphaQuery);
-	$request = loadLanguage($request);
+	$hyphaRequest = new HyphaRequest($hyphaQuery, $isoLangList);
+	$request = $hyphaRequest->getRequestParts();
```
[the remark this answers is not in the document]
laurensmartina: true, I'll remove it

system/core/pages.php
laurensmartina: There was an indentation error. :-)

system/core/HyphaRequest.php
laurensmartina: Good catch, fixed it.

system/core/HyphaRequest.php
laurensmartina: Agreed, I'll create a `getArgs()` method for this.

system/core/HyphaRequest.php
laurensmartina: ... sorry.. the implementation was correct.. you cannot detect the page type from the query if there is a language specified.

laurensmartina:
Updated!

matthijskooijman:
Ah, a language indicates a normal page, while a page type indicates a special type of page? I can't recall what we discussed about this anymore.

It also seems the page type is not used yet, will that be changed?

I've added you as a collaborator on this repository (as we discussed a few weeks ago), feel free to merge or push directly.

laurensmartina:
I've removed the page type to avoid confusion.

Issue #19, opened 2018-03-05
Open source logo is called from elsewhere


tammoterhark:
The logo in the footer is called from an external website. This is not preferred.

giplt:
I guess the CC logo(s) could be part of a dedicated set of icons to be packaged in the default theme like proposed in #24

tammoterhark:
Overgebracht naar hypha-improvements.

Issue #18, opened 2018-03-05
Proposal: develop possibility to choose between themes


tammoterhark:
Why?

As discussed with Harmen: we only release Hypha with one theme: default. But user may want to develop own theme. Currently this can be done by editting the file /data/hypha.css (why in data?).

I think it would be cleaner to make a new directory: /system/themes/default and put all styling in there. When user wants to develop own theme, user can either changes files in /system/themes/default or copy default to /system/themes/[new name] and then change whatever is necessary. Needed: possibility to change to new theme when logged in as admin.

Advantage: whenever user runs into trouble, choosing "default" theme in backend can be used to check if thing are still all right in default theme. So user can check if error is in theme or elsewhere.

matthijskooijman:
> Currently this can be done by editting the file /data/hypha.css (why in data?).

Because it can be edited by the user, and thus is effectively site-specific data (I guess).

One downside of the current approach is that when running from git, making changes to the CSS leaves you with modified files in the git checkout, and conflicts when the CSS is changed in the repository. Having separate user-defined themes would remove the need for modifying the builtin css, which should solve this problem too.

giplt:
Correct. All system files are (or could be) static, all user editable files are in data. System configuration data in root (/data), page content in subfolders (/data/images, /data/pages etc)

Distinction between a default theme and a user-defined theme seems to be a good approach to me.

tammoterhark:
The distinction between data and core was misinterpreted by me: I thought data was data.

Maybe an idea:

/system
/user-files
/user-files/themes (for custom themes)
/user-files/data (for pages and settings)
/user-files/data/images

matthijskooijman:
See #23 for discussion about a different structure.

tammoterhark:
Double

PR #20, opened 2018-03-05
Added logo so no need for external call

tammoterhark:
Addressing: https://github.com/PlanBCode/hypha/issues/19


matthijskooijman:
I do not really like adding (extra) separate files to the root, but I'm not sure if we currently have a better place for this. Perhaps a new directory can be created, to also hold the base CSS when #18 is implemented?

giplt:
How about placing it in /data/images?

tammoterhark:
As it is now, the image is referred to from hypha.xml, which lives in /data, so yes, the image should go in /data/images. That is where I placed it.

tammoterhark:
Te vroeg, wil nog even verder spelen.

PR #21, opened 2018-03-05
Css review

tammoterhark:
Reviewed both CSS and hypha.html and did some tidying up. let me know what you think of this.


tammoterhark:
Te vroeg, wil nog even verder spelen.

PR #22, opened 2018-03-06
Css review

tammoterhark:
OK, done for now: I would really like you all to look at this and give feedback. My main goal was to make something that looks both neutral as well as easy to look at (in order to get credibility).

Next to that I wanted to use some new things:
CSS variables
Implementation of lightweight CSS framework (chota)
Reshuffle some blocks so they are arranged more logically.

For futher steps I will need some feedback first and then some assistance, as some changes need code changes and that is beyond me.


tammoterhark:
First thing I noticed: using variables makes browsing much slower than when you use a preprocessor...

tammoterhark:
testhypha

tammoterhark:
mockup

tammoterhark:
Outdated

PR #85, opened 2018-10-09
Collected fixes and improvements

laurensmartina:


system/datatypes/festival.php
```diff
@@ -306,7 +306,7 @@ function showSignup($form = null) {
 		 * pay.
 		 */
 		function handleSignup($arg) {
-			global $hyphaUrl, $hyphaLanguage, $hyphaPage;
+			global $hyphaUrl, $hyphaContentLanguage, $hyphaPage;
```
matthijskooijman: In addition to introducing RequestContext, this commit also introduces the split between UI language and content language. This was originally a separate commit, I think. It is probably too much work to split these now (since then the original loadLanguage would need to be re-introduced and modified), but this split should probably be mentioned in the commit message.

system/languages/en.php
```diff
@@ -97,6 +97,7 @@
 		"changed-css" => " edited the website's CSS layout",
 		"changed-html" => " edited the website's HTML structure",
 		"changed-page" => " edited the following page: ",
+		"deleted-page" => " deleted the following page: ",
```
matthijskooijman: These changes add strings that are not used until the next commit, so they would make more sense in the next commit?

system/core/base.php
```diff
@@ -482,6 +482,15 @@ function hypha_addPage($type, $language, $name, $private) {
 		return hypha_setPage($newPage, $language, $name, $private);
 	}
 
+	function hypha_deletePage($id) {
```
matthijskooijman: Should this have a comment saying it only modifies hypha.xml and you should usually use deletePage instead?

system/core/base.php
```diff
@@ -482,6 +482,15 @@ function hypha_addPage($type, $language, $name, $private) {
...
+		global $hyphaXml;
+		$hyphaXml->requireLock();
+		$targetPage = hypha_getPageById($id);
+		if ($targetPage instanceof HyphaDomElement) {
+			$hyphaXml->getElementsByTagName('pageList')->Item(0)->removeChild($targetPage);
```
matthijskooijman: Should we *really* delete pages? Or mark them as deleted and allow an admin to view a list of deleted pages? Probably a separate discussion that should not block this merge, though.

index.php
```diff
@@ -45,7 +45,10 @@
 
 	foreach (scandir("system/core/") as $_file) if (substr($_file, -4) == '.php') require_once("system/core/" . $_file);
 
+	$pageTypeIndex = count(get_declared_classes()) -1;
 	foreach (scandir("system/datatypes/") as $_file) if (substr($_file, -4) == '.php') include_once("system/datatypes/" . $_file);
+	$hyphaPageTypes = [];
+	foreach (array_slice(get_declared_classes(), $pageTypeIndex) as $class) if (in_array('Page', class_parents($class)) && (new \ReflectionClass($class))->isInstantiable()) $hyphaPageTypes[] = $class;
```
matthijskooijman: Interesting approach to autodetect datatype classes. Not sure if I really find it elegant or hacky, though :-p

In any case, this change would be better in a separate commit, or should at least be mentioned by the commit message.

system/core/base.php
```diff
@@ -528,6 +528,7 @@ function hypha_setPage($page, $language, $name, $private) {
 				$page->appendChild($newLanguage);
 			}
 		}
+		$private = in_array($private, ['true', '1', 'on']) ? 'on' : 'off';
```
matthijskooijman: Is this "is private" check (or rather, "normalize boolean" check) something to put in a util function (which should probably return bool, not 'on'/'off'? Probably something that is fine for a new commit after merging this PR, though.

system/core/pages.php
```diff
@@ -133,7 +133,7 @@ function loadPage(RequestContext $O_O) {
 			}
 			$_node = hypha_getPage($O_O->getContentLanguage(), $_name);
 
-			if ($_node) {
+			if ($_node && (!in_array($_node->getAttribute('private'), ['true', '1', 'on']) || isUser())) {
```
matthijskooijman: Doesn't this mean the user gets a 404 when they're not logged in, rather than a message saying they should login? This is better from a strict security perspective, but leakage of private page names does not seem like a big issue here and could be less important than usability?

system/core/base.php
```diff
@@ -664,7 +665,10 @@ function hypha_indexPages($language) {
 		// get list of available pages and sort alphabetically
 		foreach(hypha_getPageList() as $page) {
 			$lang = hypha_pageGetLanguage($page, $language);
-			if ($lang) if (isUser() || ($page->getAttribute('private')!='on')) $pageList[] = $lang->getAttribute('name').($page->getAttribute('private')=='on' ? '&#;' : '');
+			if ($lang) if (isUser() || ($page->getAttribute('private')!='on')) {
+				$pageList[] = $lang->getAttribute('name').($page->getAttribute('private')=='on' ? '&#;' : '');
+				$pageListDatatype[$lang->getAttribute('name')] = $page->getAttribute('type');
```
matthijskooijman: This $pageListDatatype variable is introduced, but not used anywhere. It also seems unrelated to private handling, making me suspect this line belongs in a later commit?

matthijskooijman: Also, adding this line is the only thing changed by this entire patch hunk I believe (other than whitespace changes), so the entire thing could then be moved.

system/core/base.php
matthijskooijman: I'm not really sure what kind of improvements this commit really makes. Ideally, the commit message should explain any change in behaviour, just saying "improved" is vague unless it is really obvious from a very short commit (and even then, explaining is better).

.gitignore
```diff
@@ -1,5 +1,4 @@
 .idea
-/system/wymeditor/
```
matthijskooijman: This commit needs a better commit message too. It does not really add WYMEditor, since we were already using that. It does add the *source* or WYMEditor, but also makes a lot of other changes to how WYMEditor works. Stating these changes explicitly would be better. Also, please mention the version of WYMEditor that is added, for future reference (version number and/or git hash), and whether any changes were made to the upstream wymeditor code (if so, adding the code verbatim first and then modifying the code in a separate commit would be preferred).

system/core/base.php
```diff
@@ -328,7 +328,7 @@ function hypha_getUserById($id) {
 		$username - identifier for the user
 	*/
 	function hypha_getUserByName($username) {
-		foreach(hypha_getUserList() as $user) if ($user->getAttribute('username') == $username) return $user;
+		foreach(hypha_getUserList() as $user) if (strtolower($user->getAttribute('username')) == strtolower($username)) return $user;
```
matthijskooijman: This also means that no two users with usernames only differing in case can ever exist. Does user registration use this same function to guarantee this?

system/core/document.php
```diff
@@ -49,7 +49,15 @@ public function __construct($file = false, $base_url = false) {
...
 				$contents = '<html><head></head><body></body></html>';
+			$previousSetting = libxml_use_internal_errors(true);
 			$this->loadHTML('<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">'.$contents);
+			$map = [LIBXML_ERR_WARNING => E_PARSE, LIBXML_ERR_ERROR => E_WARNING, LIBXML_ERR_FATAL => E_ERROR];
+			foreach (libxml_get_errors() as $error) {
+				if (801 !== $error->code) {
+					error_log($error->message, $map[$error->level]);
+				}
+			}
+			libxml_use_internal_errors($previousSetting);
```
matthijskooijman: This seems unrelated to styling?

system/core/events.php
```diff
@@ -451,7 +451,7 @@ function notifyTimer() {
 					var next = msg.nextSibling;
 					if (msg.nodeType === 1) {
 						if(!msg.hasAttribute('time')) msg.setAttribute('time', now);
-						else if (now - msg.getAttribute('time') > 5000) msg.parentNode.removeChild(msg);
+						else if (now - msg.getAttribute('time') > 10000) msg.parentNode.removeChild(msg);
```
matthijskooijman: This seems unrelated to styling?

.gitignore
```diff
@@ -4,3 +4,4 @@
 /data/hypha.xml
 /data/digest
 /data/hypha.stats
+/mockup/node_modules
```
matthijskooijman: This commit adds third-party sources. It would be good if the commit message lists the upstream locations and versions of these.

index.php
```diff
@@ -36,40 +36,29 @@
...
 		Additional scripts and data are included.
 		- *core* Include core scripts
 		- *dataypes* Include modules for available datatypes: textpage, mailinglist, blog et cetera
 		- *languages* Compile a list of available user interface languages
 	*/
 
-	$_handle = opendir("system/core/");
-	while ($_file = readdir($_handle)) if (substr($_file, -4) == '.php') require_once("system/core/".$_file);
-	closedir($_handle);
+	foreach (scandir("system/core/") as $_file) if (substr($_file, -4) == '.php') require_once("system/core/" . $_file);
```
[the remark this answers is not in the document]
laurensmartina: Moved to own commit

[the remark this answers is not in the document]
laurensmartina: Moved to own commit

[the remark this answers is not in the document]
laurensmartina: Moved to own commit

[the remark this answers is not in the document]
laurensmartina: Moved to own commit

index.php
```diff
@@ -83,14 +81,29 @@
 		Check is a user is logged in and load user data. Add login/logout functionality according to session login status.
 	*/
 
-	loadLanguage($hyphaRequest);
+	// Build request context
+	$O_O = new RequestContext($hyphaRequest, hypha_getDefaultLanguage());
 
-	// Load user and page and execute the posted command. The latter
+	// Set user as global variable.
+	$hyphaUser = $O_O->getUser();
```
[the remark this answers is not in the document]
laurensmartina: added @deprecated

system/core/RequestContext.php
```diff
@@ -0,0 +1,96 @@
...
+			$rootDir = $this->getRootPath();
+			foreach ($this->getDictionaryLanguageOptions() as $lang) {
+				$file = $rootDir . '/system/languages/' . $lang . '.php';
+				if (file_exists($file)) {
+					return include($file);
+				}
+			}
+
+			return [];
+		}
```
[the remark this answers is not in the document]
laurensmartina: Nice catch! Implemented cache

system/core/pages.php
```diff
@@ -89,28 +89,6 @@ function newPage() {
...
-		$language = $hyphaRequest->getLanguage();
-
-		// set wiki language. we want to store this in a session variable, so we don't loose language when an image or the settingspage are requested
-		$hyphaLanguage = $language !== null ? $language : hypha_getDefaultLanguage();
-
-		if (!isset($_SESSION['hyphaLanguage']) || $hyphaLanguage != $_SESSION['hyphaLanguage']) {
-			session_start();
-			$_SESSION['hyphaLanguage'] = $hyphaLanguage;
-			session_write_close();
-		}
```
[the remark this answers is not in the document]
laurensmartina: Added extra line in commit message

system/languages/en.php
laurensmartina: Moved to other commit

system/core/base.php
laurensmartina: From a scope perspective I think it's a bit weird that `hypha_deletePage` "knows" about `deletePage`. I think the readme should contain a chapter containing datatypes and this information should be placed there.

system/core/base.php
laurensmartina: Good idea! This can be implemented if needed.

index.php
laurensmartina: I understand that it looks a bit hacky but I do believe this to be less error prone.

I've put it in a separate commit.

system/core/base.php
laurensmartina: Moved $pageListDatatype to other commit

system/core/base.php
laurensmartina: Agreed, a util function would be better.

I've updated the commit message to explain the change.

system/core/pages.php
laurensmartina: When, when a not-logged-in user tries the access a private page they will see a login-to-view message.

system/core/base.php
laurensmartina: Yes, no two users with  usernames only differing in case can ever exist.
Yes, the user registration uses this same function to guarantee this.

system/core/document.php
laurensmartina: The `$file` can contain non-strict html that would cause errors, therefore those errors are ignored. In my opinion the non-strictness is only for styling purposes.

system/core/events.php
laurensmartina: Changed commit message to "improved UX".

.gitignore
laurensmartina: This actually ignores third-party sources.

.gitignore
laurensmartina: The commit is split as suggested, the commit message is updated.

index.php
```diff
@@ -61,10 +61,8 @@
 	$hyphaUrl = $hyphaRequest->getRootUrl();
 
 	// Shortcut for direct file requests
-	if ($hyphaQuery == 'data/hypha.css')
-		serveFile($hyphaQuery, false);
-	if (startsWith($hyphaQuery, 'system/wymeditor'))
-		serveFile($hyphaQuery, 'system/wymeditor');
+	if ($hyphaQuery == 'data/hypha.css') serveFile($hyphaQuery, false);
+	if (startsWith($hyphaQuery, 'system/wymeditor')) serveFile($hyphaQuery, 'system/wymeditor');
```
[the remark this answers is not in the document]
laurensmartina: Agreed. I'll do that next time.

system/core/pages.php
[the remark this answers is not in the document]
matthijskooijman: Did you push your changes? I can't see anything in the commit message here: https://github.com/PlanBCode/hypha/pull/85/commits/837b9eb27d98d71c2d745fdc9b09497c81e78f8e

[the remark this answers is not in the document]
laurensmartina: I think I reverted the commit when something when wrong and I forgot to put back the commit message. Now again I've added an extra detail in the commit message.

system/core/base.php
matthijskooijman: I created #88 for this.

system/core/pages.php
matthijskooijman: I've added a bit more to the commit message for this change:

>    Additionally, the check for a private page now happens in loadPage,       
>    rather than in the page's show(). This helps to make sure that the flag               
>    is always checked, and makes sure a 404 is returned.

system/core/document.php
matthijskooijman: I've moved the libxml change to its own commit, and clarified in the code that this only ignores unknown tag errors, not *all* errors as I previously thought when quickly reading the code.

.gitignore
matthijskooijman: I meant the commit this change is in, not this particular change (I just took the first change in the commit, since you can't comment on a commit anymore in github). However, I see now that this is not really clear.

In any case, I meant files like normalize.css. I've now found the source for this file and put it in the commit message. I've also removed chota.css and cclogo.png, both of which seemed unused (and chota.css was in data/ rather than data/css, which seemed odd).

For the dosis font, I couldn't find a canonical source. Perhaps we should clarify the source along with the license (see #89).

system/core/base.php
```diff
@@ -41,7 +41,9 @@
 	include_once ('language.php');
 	include_once ('crypto.php');
 
-	if (!is_file('data/hypha.xml')) die('serious error: missing system file hypha.xml');
+	// jump to installation procedure when hypha.xml is missing
+	if (!is_file('data/hypha.xml')) header('Location: hypha.php');
+
```
matthijskooijman: I've moved this check into index.php, so it can incorporate the root url. This fixes an infinite redirect when trying to access e.g. /en/SomePage without a datafile, which would redirect to /en/hypha.php, which would keep redirecting to itself.

matthijskooijman:
I've made some more small changes (as commented) and force pushed to this branch. As far as I'm concerned, this PR is ready for merging. I'll leave it to you guys tonight to do a final review and do the actual merge.

tammoterhark:
Wie durft er op de knop te drukken?

matthijskooijman:
Ik! Gemerged :-D

tammoterhark:
Top. Great step for mankind!

PR #100, opened 2018-11-08
Introduced search on text pages & articles

tammoterhark:
By Bert Zuurke


index.php
```diff
@@ -164,3 +166,23 @@
...
+	* createSearch
+	* makes the html for the search bar
+	*/
+function createSearch()
+{
+  $search = __('search');
+	$searchbar= <<<EOF
+	<!-- HTML for SEARCH BAR -->
+	<span >
+	<input id="mysearchPattern" type="text" class="searchinput" placeholder="$search ..." name="q" size="21" maxlength="120" style="width: 300px; ">
```
matthijskooijman: Should we change 'mysearchPattern' here? It looks like it comes from an example? Perhaps just "searchPattern"?

index.php
```diff
@@ -164,3 +166,23 @@
...
+	<button style="margin-left: 15px; margin-right: 15px;" class="searchbutton" type="search"
```
matthijskooijman: Should this styling be here, or in the CSS file? Same for styling on input field above.

system/core/HyphaRequest.php
```diff
@@ -16,6 +16,7 @@ class HyphaRequest {
 		const HYPHA_SYSTEM_PAGE_SETTINGS = 'settings';
 		const HYPHA_SYSTEM_PAGE_UPLOAD = 'upload';
 		const HYPHA_SYSTEM_PAGE_CHOOSER = 'chooser';
+    const HYPHA_SYSTEM_PAGE_SEARCH = 'search'; // bz zoek
```
matthijskooijman: The comment should be removed.

system/core/HyphaRequest.php
```diff
@@ -121,6 +122,7 @@ public static function getSystemPages() {
 				HyphaRequest::HYPHA_SYSTEM_PAGE_SETTINGS,
 				HyphaRequest::HYPHA_SYSTEM_PAGE_UPLOAD,
 				HyphaRequest::HYPHA_SYSTEM_PAGE_CHOOSER,
+        HyphaRequest::HYPHA_SYSTEM_PAGE_SEARCH, // bzhelp
```
matthijskooijman: The comment should be removed.

system/core/pages.php
```diff
@@ -195,6 +195,15 @@ function loadPage(RequestContext $O_O) {
...
 				}
 				break;
+        // hier invoegen
+        case HyphaRequest::HYPHA_SYSTEM_PAGE_SEARCH: //bz search
+          $hyphaHtml->writeToElement('pagetitle',__('searchresults')); //add searchtitle to languages
+          $hyphaPage = new searchPage($args);
+           $contentLanguage = $O_O->getContentLanguage();
+           $hyphaHtml->writeToElement('langList', hypha_indexLanguages('', $contentLanguage));
+        break;
+        // einde search
```
matthijskooijman: I would also drop this comment

system/core/pages.php
```diff
@@ -195,6 +195,15 @@ function loadPage(RequestContext $O_O) {
 					exit;
 				}
 				break;
+        // hier invoegen
+        case HyphaRequest::HYPHA_SYSTEM_PAGE_SEARCH: //bz search
```
matthijskooijman: This comment (and on the next line) can also be dropped?

system/core/pages.php
```diff
@@ -195,6 +195,15 @@ function loadPage(RequestContext $O_O) {
...
+          $hyphaHtml->writeToElement('pagetitle',__('searchresults')); //add searchtitle to languages
```
matthijskooijman: This sets 'pagetitle', but this element does not seem to be defined anywhere? At least no in my git master version?

system/datatypes/searchPages.php
```diff
@@ -0,0 +1,257 @@
...
+		global $isoLangList;
+		global $O_O;
+		$pattern = str_replace('_'," ",$pattern);
+		if ($O_O->isUser()) {
+			$includePrivatePages = true;
+		} else {
+				$includePrivatePages = false;
+			}
+        $context = 100; // determines the length of the context around the pattern in case text is found
+		$responce = hyphaSearchPattern($pattern,$includePrivatePages,$context);
```
matthijskooijman: Typo, should be $response

system/datatypes/searchPages.php
```diff
@@ -0,0 +1,257 @@
...
+* private - true   on  = include private pages (user is logged on)
+*           false  off = exclude private pages
+* Return:
+* html formatted list of found file references
+* clickable titel
+* text fragment with search pattern in red
+*
+*/
+
+function hyphaSearchPattern($spattern,$bprivate=false, $icontextlength=100)
```
matthijskooijman: Wouldn't it be better to make this function (and the rest below) methods of the `searchPage` class? That reduces chances of name collision and better indicates the structure of things?

system/datatypes/searchPages.php
```diff
@@ -0,0 +1,257 @@
...
+{
+$hyphaData = "./data/";
+$hyphaPages = "./data/pages/";
+if (strlen($spattern) == 0) return "No pattern specified";
+$context = max ($icontextlength,strlen($spattern));
+$_htmlResult =  "";
+$dom = new DOMDocument();
+$dom->load($hyphaData . 'hypha.xml');
+$pglstnr = 0;
+foreach($dom->getElementsByTagName('pageList') as $pagelist)
```
matthijskooijman: The code in this function (above, and a bit below) duplicates code also found elsewhere. I believe it should be possible to use the API for that here. For inspiration, the index page is probably good to look at, which also iterates over all pages: https://github.com/PlanBCode/hypha/blob/master/system/core/base.php#L686

matthijskooijman: This also applies to other code below, but I do not have examples handy for those (and no more time right now).

system/datatypes/searchPages.php
Bezuur: done

system/core/pages.php
Bezuur: done

system/core/pages.php
Bezuur: done

system/core/HyphaRequest.php
Bezuur: Done

system/core/document.php
```diff
@@ -257,7 +256,7 @@ class HTMLForm {
 
 			Parameters:
 			$form - A DOMWrap\Element that contains the form
-			        fields.
+							fields.
```
matthijskooijman: Here, the indentation was correct,  but is now no longer. This is is a bit of a special line, since it consists of two types of indentation: First a number tabs that indent the entire block. These tabas should match the preceding lines. Then, there should be a number of spaces, to make sure that "fields" lines up with "A DOMWrap". This second type should be spaces rather than tabs, since proper alignment depends on the tab width used for display (unlike the first type of indentation, where the only alignment needed is with other lines also indented by tabs, so you can change the tab display width to change the look, without messing up any alignment).

This is one of the reasons I often use space-only indentation in my code, but with some consideration, this tab-based (mixed with spaces in cases like this) can also work fine.

system/datatypes/searchPages.php
```diff
@@ -0,0 +1,258 @@
...
+
+		function showsearchhelp($pattern) {
+			// the command should contain a type and pattern
+			// e.g. /search/test/welcome
+			// if the type is missing this function is called and gives this message
+			// due to the fact that search is always activated via the search button this function is seldom activated
+		global $uiLangList;
+		global $isoLangList;
+		$response = "Error Internal error:illegal call to search, specify search type";
+		$this->html->writeToElement('main', $response);
```
matthijskooijman: These lines still look underindented.

system/core/document.php
```diff
@@ -34,8 +35,6 @@ function registerPostProcessingFunction($func) {
 	*/
 
 	class HTMLDocument extends DOMWrap\Document {
-		const XML_HTML_UNKNOWN_TAG = 801;
```
matthijskooijman: Why move this constant into global scope?

matthijskooijman:
Superseded by #159 .

Issue #91, opened 2018-10-25
Wikifying of links no longer works


matthijskooijman:
Originally, when creating a link to an internal page, it would be wikified when saving (meaning it would be stored as a hypha: link internally, and then would be rendered back into a normal url when viewing the page. This made sure that when the page was renamed, any links would keep working, and the link text would be updated too.

Since merging #85, this has broken. I suspect that the new wymeditor code perhaps just not calls wikify at all, but I haven't investigated.

A related enhancement (that is already in place on dewar.nl, but not in a superclean way) would be to only touch the link text when it is still identical to the page title. This allows using a different link text when referring to a page, but when the page title is used, it will still be updated when the page is renamed.

matthijskooijman:
> A related enhancement (that is already in place on dewar.nl, but not in a superclean way) would be to only touch the link text when it is still identical to the page title. This allows using a different link text when referring to a page, but when the page title is used, it will still be updated when the page is renamed.

Turns out I already implemented this in 08ecabbd7a08ecabbd7ac352e32dfaa8da1949fc85a377756cc352e32dfaa8da1949fc85a377756c

Also, it seems that this was not broken by #85, but I suspect 148fc9203d86bdf9febbfb2e239028442c417466 or c091abd66f1c4864dd317237e93766e702a99ed9 broke this. Working on a fix now.

PR #147, opened 2018-11-28
Fix wikifying of links

matthijskooijman:
This fixes the wikifying of links, as reported in #91. Additionally, this contains a few small fixes I came across when looking for this bug.


matthijskooijman:
Updated to latest master and merged :-)

PR #149, opened 2018-11-29
Article user interface enhancements

Bezuur:
Please review the changes.


system/datatypes/peer_reviewed_article.php
```diff
@@ -567,6 +567,7 @@ private function updateDiscussionsContainer($type, $discussionsContainer) {
 			$list = $this->xml->createElement('ul');
 			/** @var HyphaDomElement $reviewCommentContainer */
 			$reviewCommentContainer = $this->xml->createElement('div');
+			$_discussionnumberlabel = __('art-title') . ":" . $_discussionsubject . '(nr. ' . $_discussionnumber . ')';
```
matthijskooijman: Should there be a space before `'(nr` here?

system/datatypes/peer_reviewed_article.php
```diff
@@ -757,7 +758,7 @@ public function discussionAction($type) {
 			$form->validateEmailField(self::FIELD_NAME_DISCUSSION_COMMENTER_EMAIL);
 		}
 		$form->validateRequiredField(self::FIELD_NAME_DISCUSSION_COMMENT . '/new_' . $type);
-
+		$form->validateRequiredField(self::FIELD_NAME_DISCUSSION_SUBJECT);
```
matthijskooijman: Did you intend to remove the empty line here? It seems useful to keep it, to separate the validation from the rest of the processing.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1048,12 +1051,27 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
 		$comment = __('art-comment');
 		$new = $discussion === null;
 		$commentFieldName = self::FIELD_NAME_DISCUSSION_COMMENT . '/' . ($new ? 'new_' . $type : $discussion->getId());
-		$html = <<<EOF
+		if (strpos($commentFieldName, 'new_review') > 0) {
```
matthijskooijman: Even though using strpos here works and is concise, it does not seem a good practice to me. I believe you want to say `if ($new && $type == 'review')` here, so probably best to just write that down. IMO that gives more readable code, and prevents problems later on when changing the field names without realizing this check must also be changed.

It is probably also *faster*, since strpos has to scan through the entire string until it finds the string you look for.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1048,12 +1051,27 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
...
+EOF;
+}
+else {
+	$html = <<<EOF
+	<div class="new-comment $type">
+	<div>
+		<strong><label for="$commentFieldName"> $comment </label></strong><textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
+	</div>
 EOF;
+}
```
matthijskooijman: The indentation of this else block seems wrong? I suspect you aligned it with the EOF, but it's probably better to align it with the original if?

system/datatypes/peer_reviewed_article.php
```diff
@@ -1048,12 +1051,27 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
 		$comment = __('art-comment');
 		$new = $discussion === null;
 		$commentFieldName = self::FIELD_NAME_DISCUSSION_COMMENT . '/' . ($new ? 'new_' . $type : $discussion->getId());
-		$html = <<<EOF
+		if (strpos($commentFieldName, 'new_review') > 0) {
+			$_discussionLabel = __('art-start-discussion');
+			$_discussionShortTitle = __('art-title');
+
+			$html = <<<EOF
```
matthijskooijman: This (and below) duplicates the entire html EOF block, so introduces quite some duplicate lines of (HTML) code. I wonder if it would be better to split the EOF block into separate lines instead, so you can always add the enclosing div tags and comment textarea, and make just the title stuff optional.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1048,12 +1051,27 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
...
 			<div class="new-comment $type">
 			<div>
-				<strong><label for="$commentFieldName"> $comment </label></strong><textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
-			</div>
+				<strong><label for="$commentFieldName">$_discussionLabel</label></strong>
+				$_discussionShortTitle<br><input type="text" name="subject"><br>
+				$comment<br>
+				<textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
```
matthijskooijman: Is the label correct here? I would think the label is about the new comment field, so it should contain `$comment`, not `$_discussionLabel`. Also, I guess the `$_discussionShortTitle` should also be placed in a label tag, with an appropriate `for` attribute.

system/languages/en.php
```diff
@@ -307,6 +307,10 @@
 		"art-add-comment" => "add comment",
 		"art-comment" => "comment",
 		"art-comments" => "comments",
+		"art-title" => "title",
+		"art-continue-discussion" => "contribute",
```
matthijskooijman: `contribute` seems weird here, it does not match the string label, nor the Dutch translation. Shouldn't this just be `Continue discussion`?

matthijskooijman:
Oh, one more thing: The commit message " adddifferent style and a label to closed discussions " is missing a space and a capital letter.

data/hypha.html
```diff
@@ -33,6 +32,7 @@ <h1 id="pagename"></h1>
             <div id="langList"/>
           </div>
           <div id="main" />
+          <div id="pageEndCommands"></div>
```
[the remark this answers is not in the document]
matthijskooijman: Did you see the commit message for this change?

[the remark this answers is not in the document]
tammoterhark: Yes, now I did ... OK.

data/hypha.css
```diff
@@ -597,6 +597,12 @@ input {
  background-color:  var(--menulinkhoverback);
 }
 
+div.comment-subject {
+  font-style: italic;
```
tammoterhark: You emphasized this title in three ways:

- make it larger
_ make it bolder
- add italic

I would prefer just to make it larger, leave out the others.

system/datatypes/peer_reviewed_article.php
```diff
@@ -564,13 +567,16 @@ private function updateDiscussionsContainer($type, $discussionsContainer) {
 			$list = $this->xml->createElement('ul');
 			/** @var HyphaDomElement $reviewCommentContainer */
 			$reviewCommentContainer = $this->xml->createElement('div');
+			$reviewTitle = $this->getXml()->createelement('div',$_discussionnumberlabel);
+			$reviewTitle->attr('class','comment-subject');
 			$class = 'review-comment-wrapper collapsed';
 			foreach (['blocking' => $blocking, 'resolved' => $resolved, 'closed' => $closed] as $name => $isTrue) {
```
tammoterhark: As a result of this (?) the title is now shown as:

title:new title(nr. 2)

I would prefer to do it as follows:

#2 new title

So leave out the label ("title"), because it is obvious. Introduce the number at frond, because a number may also be part of the comment title itself.

tammoterhark: `<h3 class="comment-title"><span class="comment-number"># 3</span> <span class="title">Dit is de titel</span></h3>`

matthijskooijman:
> This now works for the page type "Peer reviewed article". Not for other page types.

What part of "this" do you mean exactly? Most of the stuff in this commit is specific to articles. Only adding buttons to the bottom seems like it could apply to other pages, but that could be a separate improvement (i.e. I think it should not delay merging this request).

tammoterhark:
Right, I could have been more specific. Indeed, I was talking about the buttons at the bottom of the screen. Meant as an invitation to implement this at other page types too, as an enhancement. 

By no means added to block anything!

matthijskooijman:
@tammoterhark, perhaps you could create a new issue for this, and perhaps have a look at which buttons would need to be repeated for each page type?

tammoterhark:
Field for title does show up for review comments, but not for reader comment. But it seems that it is needed.

Image shows situation after submitting reader comment on published peer review article. 

commentaarzoekttitel

system/languages/en.php
Bezuur: No, I ment "contribute to **this** discussion"

matthijskooijman: In that case, perhaps the id of the string should be changed to "art-contribute" or something similar?

Bezuur:
I think that the label is a placeholder and had onlusten a Meininger for the code.
The message that replace the placeholder is to help the user to understand what to do next.
So in my opinion the wording of the placeholder and replacement need not be the same.
If you see this construction as an translator, than you have a point.


Met vriendelijke groeten
Bert Zuurke
(Verstuurd vanaf mijn iPad)

Bezuur:
We need to discuss this further.

Met vriendelijke groeten
Bert Zuurke
(Verstuurd vanaf mijn iPad)

matthijskooijman:
@Bezuur, what comment are you responding to exactly? And what do you mean with " onlusten a Meininger"?

data/hypha.html
```diff
@@ -15,7 +15,8 @@
       <div id="trail" />
       <div id="log" />
     </div>
-    <div class="container pageContainer">      
+<<<<<<< HEAD
```
matthijskooijman: W00ps, we should remove that one.

matthijskooijman:
I've moved over most of these commits to #205. Some other things I have here locally and I'll use them as part of a fix for #199. With that, I think this PR can be closed.

PR #153, opened 2018-11-30
changed short open tag in line 127 to prevent errors during installation

dianawi:


matthijskooijman:
Commit looks good. The commit message has a bit of a long first line. In particular, stating the line number is not really needed, git knows this already (also, stating the line number without the filename does not really help). If I would write a message for this commit, I would probably write something like:

> Replace short open tag with long one
>
> In recent PHP versions, short open tags are no longer enabled by
> default, so this short open tag would cause a syntax error and prevent
> hypha from running at all.

This also gives some info on why this short open tag is problematic.

Care to practice with amending a commit and force pushing to update this PR? Or shall we just merge as-is?

Issue #156, opened 2018-12-06
confirm subscription newsletter


dianawi:
1. when you subscribe to the newsletter there is no message saying you have to check your mail for a confirmation request.
Related to #29.  
2. In the confirmation email the text is unclear: bevestig de e-mail/confirm the email
3. when you confirm you go back to the sign up page, where it looks like you have to sign up again. How to fix that? give a message: 'you have successfully singed up up to the newsletter. On this page you find all the previous newsletters'? Or redirect them to the last newsletter in the list?

matthijskooijman:
The first point is similar to https://github.com/PlanBCode/hypha/issues/99#issuecomment-436954058, about public comments to articles.

dianawi:
update: there is a message saying "succesvol aangemeld" (successfully signed up) way at the bottom of the screen, where no one can see it unless they scroll down very fast. 
the other message is also there, in that same place:"Succesvol verstuurd, bevestig de e-mail" (success, confirm by email). So it seems that everything works, but is in the wrong place.

PR #157, opened 2018-12-08
Fix subscribe message

dianawi:
@matthijskooijman: I changed the position of all hyphaNotify's. This has consequences for all notifications, and may also make it easier for visitors to leave a comment. Also made the 'success' notifications green instead of red.
I would prefer a message that users have to actively close, but for now this already helps.
@tammoterhark: I have changed the notification style on hypha.css, but for De Stadsbron we have to change it in the CSS of De Stadsbron.
Fixes:#156


system/languages/en.php
```diff
@@ -248,14 +248,15 @@
...
 		"ml-invalid-command" => "invalid command",
 		"ml-inconsistent-parameters" => "the given parameters are inconsistent",
-		"ml-successfully-subscribed" => "successfully subscribed",
-		"ml-successfully-unsubscribed" => "successfully unsubscribed",
+		"ml-successfully-subscribed" => "you have successfully subscribed",
+		"ml-successfully-unsubscribed" => "you have successfully unsubscribed",
 		"ml-successfully-created" => "successfully created",
 		"ml-successfully-updated" => "successfully updated",
-		"ml-successfully-sent" => "successfully sent",
+		"ml-successfully-sent" => "You will soon receive an email with a link. Please confirm your subscription. If you do not get the email, it may have landed in your spam folder.",
```
matthijskooijman: This changes the meaning of this string significantly. I would suggest we also rename it, to something like `ml-please-confirm-by-email` or something?

system/datatypes/mailinglist.php
```diff
@@ -148,10 +148,10 @@ public function indexAction() {
 					// send email so that action can be confirmed
 					$link = $this->constructFullPath(sprintf('%s/confirm?code=%s', urlencode($this->pagename), urlencode($code)));
 					$confirmTxt = __('ml-please-confirm-email');
```
matthijskooijman: Is this `$confirmTxt` still used anywhere? If not, the corresponding string should probably be removed too.

matthijskooijman: Yup, still used in the e-mail body.

system/languages/en.php
dianawi: renamed

PR #158, opened 2018-12-12
Add editor welcome message

dianawi:
this makes it possible for users to enter their own welcome message to new mailinglist subscribers


matthijskooijman:
The commit that was added most recently changes the translated strings from "description" to "welcome message" (or something similar), but all the internals are still "description" (e.g. the field names, the string id, etc.). For consistency, I would change all of these.

The last commit has an updated commit message, but I think it should really be squashed with the previous commit, so we should do that before merging this.

PR #159, opened 2018-12-13
Search add on

Bezuur:
This enables a basic search function for Hypha


matthijskooijman:
See #100 for some earlier review comments.

PR #160, opened 2018-12-13
Help icoontje

Bezuur:
add help and information features to Hypha, featuring:
 - a general information page with all the help keywords and explanation
 - a system to add information icons to labels, pages and buttons for popup information
 - the keywords and explanation are expandable without programming


system/core/events.php
```diff
@@ -74,7 +162,7 @@ function hypha(url, cmd, arg) {
 		document.getElementById('argument').value = arg;
 		document.forms['hyphaForm'].action = url;
 		for(i=0; i<postProcessingList.length; i++) postProcessingList[i]();
-		if (cmd||arg) $(document.forms['hyphaForm']).submit();
+		if (cmd||arg) (document.forms['hyphaForm']).submit();
```
matthijskooijman: What is this change for? I believe this switches from the jquery version of `submit()` to the regular version, which IIRC are not completely identical. If this is indeed needed for the info icons, the commit message should probably explain why, and this could perhaps also be moved into a separate commit.

system/core/pages.php
```diff
@@ -195,6 +210,24 @@ function loadPage(RequestContext $O_O) {
 					exit;
 				}
 				break;
+
+				case HyphaRequest::HYPHA_SYSTEM_PAGE_HELP: //bz help
```
matthijskooijman: Indentation again.

system/core/pages.php
```diff
@@ -195,6 +210,24 @@ function loadPage(RequestContext $O_O) {
...
+				if ($args[0] == 'help') {
+							echo getButtonInfo($args);
+						exit;
+				} else {
+					$hyphaHtml->find('#pagename')->text( __('help-pagetitle')); //		"help-pagetitle" => "Help index",
+					$hyphaPage = new helpPage($args);
+					//echo "pages: 222<br>\n";
+					if (count($args) == 2) if (hypha_isLanguage($args[1])) {//echo " -pages.php 222- " . $args[1];
```
matthijskooijman: This should not stack two ifs in a single line, since now it is not really clear where the else belongs.

system/core/pages.php
```diff
@@ -195,6 +210,24 @@ function loadPage(RequestContext $O_O) {
...
+						$hyphaHtml->writeToElement('langList', hypha_helpLanguages('',$args[1]));
```
matthijskooijman: Why do you handle the language selection manually here? Does HyphaRequest not already handle that so you can just use `$O_O->getContentLanguage();` always? Or does it not handle languages for system pages, perhaps?

system/core/pages.php
```diff
@@ -534,3 +567,60 @@ function image_resize($src, $dst, $width, $height) {
 		}
 		return true;
 	}
+
+	function getButtonInfo($_args) {
```
matthijskooijman: I'm confused by this name, AFAIK it retrieves generic help/tooltip content, not just limited to buttons?

system/core/pages.php
```diff
@@ -534,3 +567,60 @@ function image_resize($src, $dst, $width, $height) {
...
+		/* 0 = buttonName
+		*  1 = buttonName
+		*  2 = language (or null)
+		*/
+		Global $O_O;
+		$_button = "";
```
matthijskooijman: Why use _ as a prefix here? These are often used to mark values as "do not use externally", but since these are local variables, they are already confined to this function only?

system/core/pages.php
```diff
@@ -534,3 +567,60 @@ function image_resize($src, $dst, $width, $height) {
...
+		$_language = "";
+
+		if (sizeof($_args) > 1) {$_button = $_args[1];} else {$_button = "undifined";}
+		$a = array("%20","%22");
+		$b = array(" "," ");
+	  $_button = str_replace($a,$b,$_button);
```
matthijskooijman: What does this replacement do? Replace escaped spaces (%20) with doublequotes (%22)? And one kind of spaces with another? This code can really use a comment to explain what it is doing and why.

system/core/pages.php
```diff
@@ -534,3 +567,60 @@ function image_resize($src, $dst, $width, $height) {
...
+		if (sizeof($_args) > 2) {$_language = $_args[2];} else {$_language=$O_O->getInterfaceLanguage();} //$_language=$O_O->getContentLanguage(); //DefaultLanguage();
+		$info = hypha_searchhelp($_button,$_language);
+		return  $info ; // "AJAX - Help voor <br>" . $_button . "<br> in de taal: ". $_language;
+	}
+
+	function hypha_isLanguage($id) {
+		$pageLangList = array("nl","en","de"); // replace by hypha language list
```
matthijskooijman: This code is unfinished. IIRC HyphaRequest already has a similar check, which could be reused (or moved here, and reused by HyphaRequest). In any case, the addition of this function would be better placed in a separate commit, since it is not tied to the tooltip stuff (this function could very well be used by other code too).

system/datatypes/helpPage.php
```diff
@@ -0,0 +1,194 @@
...
+
+function helpinfo($subject,$language){
+$info =  "HELPPAGES.PHP HELPINFO: argument subject ". $subject . " in de taal ". $language . "<br>\n";
+//	$info = hypha_searchhelp($subject, $language);
+	return $info;
+}
+
+function hypha_searchhelp($subject,$language = 'en') {
+	$subject = str_replace("%20", " ",$subject);
+	if (!$language) $language='en';
```
matthijskooijman: Same comments about the language defaults as for `hypha_helpIndex`

system/datatypes/helpPage.php
```diff
@@ -0,0 +1,194 @@
...
+}
+
+function helpinfo($subject,$language){
+$info =  "HELPPAGES.PHP HELPINFO: argument subject ". $subject . " in de taal ". $language . "<br>\n";
+//	$info = hypha_searchhelp($subject, $language);
+	return $info;
+}
+
+function hypha_searchhelp($subject,$language = 'en') {
+	$subject = str_replace("%20", " ",$subject);
```
matthijskooijman: If this replacement is needed, it should probably happen elsewhere. Urlencoding is associated with parsing and generating URLs, while this code is low-level database-access code, which should not have to know about URLs.

system/core/events.php
```diff
@@ -58,7 +58,95 @@ function addEventHandler($html) {
...
+			// check if id has an entry in clicked
+			for(var key in clicked) {
+				if (key == id)
+				 if (clicked[key] == true){hideinfod(id); return;}
+			 }
+			 clicked[id] = true;
+			 	/* get the help info from the hypha server */
+					var pageLanguage = document.getElementById('langList')
+					var language = pageLanguage.getElementsByClassName('selected');
+		      url = "help/help/" + subject + "/" + language[0].innerHTML;
```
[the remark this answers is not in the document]
Bezuur: Yes, because the help language is different from the page and system languages. The help information can differ from the system and page information.

system/core/events.php
```diff
@@ -58,7 +58,95 @@ function addEventHandler($html) {
...
+						info.top = x;
+						info.left = y;
+		      } );
+		      /* help info added, position the popup */
+		    info.style.display = "block";
+		    info.position = "absolute";
+				info.top = x;
+				info.left = y;
+				var text = info.getElementsByTagName("p").item(0);
+				text.innerHTML = this;
```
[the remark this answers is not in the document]
Bezuur: I 'm working on a function to place the info popup based on a screen model. So this will change in a future commit.

system/core/events.php
Bezuur: In my context the $(comment) triggers an error.
not reproducable, so added the $ again.

system/core/pages.php
Bezuur: The language for the help system differs from the languages for the pages. When the help index page is shown, the languages in the languages list are the languages from the help database.

matthijskooijman:
Superseded by #178

PR #163, opened 2018-12-29
Mailinglist UI improvements

giplt:
A number of small changes to improve the workflow of sending out a message to a mailing list

### DIFF
--- a/system/datatypes/mailinglist.php
+++ b/system/datatypes/mailinglist.php
@@ -31,6 +31,7 @@ class mailinglist extends Page {
 	const FORM_CMD_LIST_DELETE = 'delete';
 	const FORM_CMD_MAILING_SAVE = 'save';
 	const FORM_CMD_MAILING_SEND = 'send';
+	const FORM_CMD_MAILING_TEST_SEND = 'test_send';
 
 	const MAILING_STATUS_DRAFT = 'draft';
 	const MAILING_STATUS_SENT = 'sent';
@@ -229,7 +230,7 @@ class mailinglist extends Page {
 				$date = '';
 			}
 			$row->addCell($date);
-			$buttons = $this->makeActionButton(__('read'), $mailing->getId());
+			$buttons = $this->makeActionButton(__('view'), $mailing->getId());
 			if (self::MAILING_STATUS_DRAFT == $status) {
 				$buttons .= $this->makeActionButton(__('edit'), $mailing->getId() . '/edit');
 			}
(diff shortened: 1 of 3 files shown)


system/datatypes/mailinglist.php
```diff
@@ -229,7 +230,7 @@ private function displayArchive() {
 				$date = '';
 			}
 			$row->addCell($date);
-			$buttons = $this->makeActionButton(__('read'), $mailing->getId());
+			$buttons = $this->makeActionButton(__('view'), $mailing->getId());
```
matthijskooijman: This change does not seem to be mentioned in the commit message? I suspect it would be a separate commit even?

system/datatypes/mailinglist.php
```diff
@@ -537,6 +539,21 @@ private function showMailingAction($mailingId) {
...
+			} catch (\Exception $e) {
+				$msg = $e->getMessage();
+				$msgType = 'error';
+			}
+			
+			// goto view page with notification
+			notify($msgType, $msg);
+			return ['redirect', $this->constructFullPath($this->pagename . '/' . $mailingId)];
+		}
+
```
matthijskooijman: There is quite some duplication between the test and regular send case. Perhaps these could be unified a bit more?

system/datatypes/mailinglist.php
```diff
@@ -770,6 +782,22 @@ private function createOrUpdateMailing(array $data) {
...
+		$senderName = $this->getDoc()->getAttribute('sender-name');
+		$senderEmail = $this->getDoc()->getAttribute('sender-email');
+		
+		/** @var HyphaDomElement $mailing */
+		$mailing = $this->xml->document()->getElementById($mailingId);
+		
+		$message = '<p><a href="' . $linkToMailing . '">' . __('ml-if-unreadable-use-link') . '</a></p>';
+		$message .= $mailing->getHtml();
+		$this->send($mailing->getAttribute('subject'), $message, [$email], $senderEmail, $senderName);
```
matthijskooijman: Same here: This code is duplicated between test send and regular send, so it would be very easy now to update one and not the other, making the test send non-representative. I can imagine merging these two functions (with a test boolean argument), or perhaps moving the duplicate code that generates (and perhaps sends) each message into its own function.

system/languages/en.php
```diff
@@ -238,13 +238,15 @@
...
+		"ml-test-send" => "test-send",
 		"ml-view-addresses" => "view addresses",
 		"ml-date-format" => "d/m/Y",
 		"ml-received-by" => "received by",
+		"ml-sure-to-send" => "Are you sure to send off this mailing to [[count]] recipient(s)?",
 
 		// mailing list - response
 		"ml-no-sender" => "no sender, please edit the mailing list",
-		"ml-mail-has-invalid-status" => "de mail heeft een onjuiste status",
+		"ml-mail-has-invalid-status" => "the mail has an invalid status",
```
matthijskooijman: This also seems a separate change.

system/languages/nl.php
```diff
@@ -154,7 +154,7 @@
 		"general-info" => "algemene informatie",
 		"lineup" => "lineup",
 		"timetable" => "blokkenschema",
-		"view" => "view",
+		"view" => "bekijk",
```
matthijskooijman: This one as well (can probably be in a commit together with the other translation fix above)

giplt:
Just split this PR into 4 separate ones: #165, #166, #167, #168.

Issue #140, opened 2018-11-26
send test newsletter to a few email addresses


dianawi:
Now there are sometimes bugs in the newsletter, like bad links, that you can only discover when the newsletter in in the email. In the preview it works, but in the email it does not.
related to #141 
possible solution: test button which gives a field where you can select users to send the test to, and additionally add an email address.

giplt:
Implemented in PR #166

PR #169, opened 2018-12-29
Mailinglist improvements

giplt:
A number of commits to improve the work flow of sending out a mailinglist message

### DIFF
--- a/system/datatypes/mailinglist.php
+++ b/system/datatypes/mailinglist.php
@@ -32,6 +32,7 @@ class mailinglist extends Page {
 	const FORM_CMD_LIST_DELETE = 'delete';
 	const FORM_CMD_MAILING_SAVE = 'save';
 	const FORM_CMD_MAILING_SEND = 'send';
+	const FORM_CMD_MAILING_TEST_SEND = 'test_send';
 
 	const MAILING_STATUS_DRAFT = 'draft';
 	const MAILING_STATUS_SENT = 'sent';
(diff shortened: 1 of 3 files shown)


matthijskooijman:
I fixed a few more things and updated this PR. I changed:
 - I rebased on top of the latest master.
 - I added a little more explanation to the first commit, explaining why the send button was removed from the editor.
 - I added a "Fixes #140" to the relevant commit message.
 - I added a commit to fix a string that was broken by #156.
 - I fixed the test sending of mails. In the latest refactor, the recipient address was no longer passed to `testSendMailing()`, so nothing was sent out. I looked back at #163, which used `$this->getArg(2)` as the recipient address, but AFAICS that only looks at the URL, not at the POST arguments. Instead, the recipient address is accessible through `$_POST['argument']`, which is what I used now. That makes me wonder if #163 ever worked?

With these changes, I think this can be merged. Some things that might be improved in the future:
 - Right now, the test mail is subtly different from the read e-mail (the view online and unsubscribe links are missing). It might be good to include those too (but with a dummy unsubscribe url).
 - It might be useful to somehow mark the test e-mail as such (e.g. an addition or prefix to the subject perhaps?)

giplt:
> With these changes, I think this can be merged. Some things that might be improved in the future:
> - Right now, the test mail is subtly different from the read e-mail (the view online and unsubscribe links are missing). It might be good to include those too (but with a dummy unsubscribe url).
> - It might be useful to somehow mark the test e-mail as such (e.g. an addition or prefix to the subject perhaps?)
I agree and thought of those, too. Couldn't settle on the best way for doing so however and chose to do the basic functionality first. Will add them to the issue-queue...

PR #177, opened 2019-01-06
Article more clear

Bezuur:
Solved subject: field not found.


matthijskooijman:
I've picked up the useful parts of this PR in a local branch, which I'll use as part of fixing #199. With that, I think this PR can be closed.

PR #178, opened 2019-01-08
Add help system with index and icon for popup help info

### LINKED ISSUE #112: add explanations tooltip, and something comparable for mobile use

laurensmartina:
Help info system with self positioning information popups.
The position of the icon determines the position of the information popup.
Visibility of the icons can toggled by the "?" icon.
Help index page with all the help keywords.

This fixes #112


laurensmartina:
Positioning of the icons still needs some work. My suggestion is to fix this in a next iteration.

index.php
```diff
@@ -130,6 +132,7 @@
 
 	// add hypha commands and navigation
 	$_cmds[] = '<a href="index/'.$hyphaLanguage.'">'.__('index').'</a>';
+	if ($O_O->isUser()) $_cmds[] = '<a href="help/index">'.__('help').'</a>';
```
matthijskooijman: Putting the help index in the menu will likely let users expect a well-written help text they can read from top to bottom. Since the current help index is more of a developer tool (especially since it shows the raw identifiers), I would certainly not put it in the menu like this.

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Directly set the clicked class, this also prevents the spamming of the system before the result is returned.
+	$info.addClass('clicked');
+
+	// Try to get page language from selected langlist.
+	if (!language) {
+		var $selectedLanguage = $('#langList .selected');
+		if ($selectedLanguage.length) {
+			language = $selectedLanguage[0].innerHTML;
```
matthijskooijman: This doesn't look entirely clean to me. Why check the page language at all and not just stick to the interface language?

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+
+	// Create an empty popup.
+	var $helpPopup = $('<div></div>').appendTo('body');
+
+	// Fetch popup content with Ajax and place it in the popup.
+	var url = 'help/help/' + key + (language ? '/' + language : '');
```
matthijskooijman: I think this needs some urlencoding.

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Creates an info popup.
+ *
+ * @param anchor Element, mouse click event, used to determine position of popup
+ * @param key string, the key for with to get the info
+ * @param language string
+ * @param info Element, the info button for which to make an info popup
+ */
+function toggleInfoPopup(anchor, key, language, info) {
```
matthijskooijman: The anchor is the click event. Is `anchor` a good / common name for this? Also, `info` is the clicked element, which isn't directly clear to me. Perhaps `info_icon` is more appropriate?

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+		$button.on('click', function () {
+			$info.trigger('closeMe');
+		});
+		$button.css('left', $helpPopup.css('width'));
+		$button.addClass('closebutton');
+		$button.append(document.createTextNode('X'));
+	});
+}
+
+function hypha_ajax(url, callback) {
```
matthijskooijman: Why not use jquery ajax?

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+		$textBox.addClass('notelo');
+	} else if ((y >= box.height) && ((box.width < x) && (x < sx - box.width))) {
+		$textBox.css('left', x - box.width / 3 + 'px');
+		$textBox.css('top', y - box.height - 40 + 'px');
+		$textBox.addClass('notelo down');
+	} else if ((y >= box.height) && (x > sx - box.width)) {
+		$textBox.css('left', x - box.width + 'px');
+		$textBox.css('top', y - box.height - 15 + 'px');
+		$textBox.addClass('notelo');
+	} else alert('gebied niet te bepalen');
```
matthijskooijman: This message should be English.

system/help/help.js
```diff
@@ -0,0 +1,188 @@
...
+ * you can supply additional style information too if you have it.
+ *
+ * @param text string
+ * @param fontSize number
+ * @param fontStyle string, the font style
+ * @param padding number
+ * @param border, boolean, indicating whether or not the include a 1px border
+ * @returns {{width: number, height: number}}
+ */
+function measureText(text, fontSize, fontStyle, padding, border = false) {
```
matthijskooijman: Why measure the size of the text in a separate div, then thrash it and then put the same text in the actual infobox div? The code creating this div here needs to be identical to the final div to make the prediction accurate, so might as well measure in the actual div?

system/languages/nl.php
```diff
@@ -336,6 +336,18 @@
...
+		// Keys met "help." prefix wordem in de help index opgenomen.
+		"help.welcome" => "Dit is hypha:",
+		"help.create-new-page" => "Maak een nieuwe Hypha pagina:\nKies het type\nGeef je de pagina een naam voor referentie.",
+		"help.page_name" => "Geef de pagina een naam, die gebruikt kan worden als referentie",
+		"help.languages_info" => "De pagina is beschikbaar in deze talen.\nKlik op een taal om de \"vertaling\" te zien.",
+		"help.help_languages_info" => "Klik op een andere taal.\nom de help \"vertaling\" te zien.",
+		"help.user" => "Een gebruiker kan alle artikelen zien.\nAlleen als een gebruiker admin rechten heeft kan deze gebruiker ook bijdragen aan artikelen/pagina's\nof eigen pagina's toevoegen",
+		"help.page" => "Een gebruiker kan alle artikelen zien.\nAlleen als een gebruiker admin rechten heeft kan deze gebruiker ook bijdragen aan artikelen/pagina's\nof eigen pagina's toevoegen",
+		"help.menu" => "Een gebruiker met admin rechten kan een menu samenstellen.\nDit menu vervang automatisch het bestaande menu\nEr is geen \"ga terug naar vorige versie\"\nMet de test knop kun je het menu testen voordat je het definitief maakt.",
+		"help.help" => "Dit",
```
matthijskooijman: This seems unfinished.

system/languages/nl.php
```diff
@@ -336,6 +336,18 @@
 		"art-start-review" => "start beoordeling",
 		"art-approve(s)" => "beoordeling(en)",
 		"art-unresolved-block(s)" => "onopgeloste breekpunt(en)",
+
+		"helppagetitle" => "Hulp onderwerpen",
+		// Keys met "help." prefix wordem in de help index opgenomen.
+		"help.welcome" => "Dit is hypha:",
+		"help.create-new-page" => "Maak een nieuwe Hypha pagina:\nKies het type\nGeef je de pagina een naam voor referentie.",
+		"help.page_name" => "Geef de pagina een naam, die gebruikt kan worden als referentie",
+		"help.languages_info" => "De pagina is beschikbaar in deze talen.\nKlik op een taal om de \"vertaling\" te zien.",
```
matthijskooijman: This is also shown when there is only one language for a page.

system/languages/nl.php
```diff
@@ -336,6 +336,18 @@
...
+		"help.help_languages_info" => "Klik op een andere taal.\nom de help \"vertaling\" te zien.",
+		"help.user" => "Een gebruiker kan alle artikelen zien.\nAlleen als een gebruiker admin rechten heeft kan deze gebruiker ook bijdragen aan artikelen/pagina's\nof eigen pagina's toevoegen",
+		"help.page" => "Een gebruiker kan alle artikelen zien.\nAlleen als een gebruiker admin rechten heeft kan deze gebruiker ook bijdragen aan artikelen/pagina's\nof eigen pagina's toevoegen",
```
matthijskooijman: Are these two intentionally identical?

system/languages/nl.php
```diff
@@ -336,6 +336,18 @@
...
 		"art-approve(s)" => "beoordeling(en)",
 		"art-unresolved-block(s)" => "onopgeloste breekpunt(en)",
+
+		"helppagetitle" => "Hulp onderwerpen",
+		// Keys met "help." prefix wordem in de help index opgenomen.
+		"help.welcome" => "Dit is hypha:",
+		"help.create-new-page" => "Maak een nieuwe Hypha pagina:\nKies het type\nGeef je de pagina een naam voor referentie.",
+		"help.page_name" => "Geef de pagina een naam, die gebruikt kan worden als referentie",
+		"help.languages_info" => "De pagina is beschikbaar in deze talen.\nKlik op een taal om de \"vertaling\" te zien.",
+		"help.help_languages_info" => "Klik op een andere taal.\nom de help \"vertaling\" te zien.",
```
matthijskooijman: Should there be a dot after "taal"?

index.php
```diff
@@ -144,6 +147,8 @@
 		$_cmds[] = makeLink(__('settings'), makeAction('settings', '', ''));
 		$_cmds[] = makeLink(__('logout'), makeAction($hyphaQuery, 'logout', ''));
 	}
+	$_cmds[] = '<span onclick="toggleInfoButtons(this)" class="toggleinfo">?</span>';
```
matthijskooijman: IMHO we should only offer this option if it persists across navigation to different pages. If on every url change the icons turn back on, this is probably only annoying.

system/core/pages.php
```diff
@@ -535,3 +549,24 @@ function image_resize($src, $dst, $width, $height) {
...
+			$language,
+			$O_O->getContentLanguage(),
+			$O_O->getInterfaceLanguage(),
+		];
+		$languagesToTry = array_unique($languagesToTry);
+
+		foreach ($languagesToTry as $language) {
+			$file = $O_O->getRootPath() . '/system/languages/' . $language . '.php';
+			if (file_exists($file)) {
+				$dictionary = include($file);
```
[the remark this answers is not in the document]
matthijskooijman: Looking at this more closely, I think a clean approach would be to add a `$language` param to `RequestContext::getDictionary` (and let RequestContext keep dictionaries for multiple languages cached). Then all dictionary loading happens in one place.

data/themes/default/hypha.css
```diff
@@ -740,3 +740,100 @@ th{
...
+.hyphaInfoButton:hover {
+    background-color: #fff;
+    color: #00f;
+}
+.hyphaInfoButton.clicked {
+    background-color: yellow;
+}
+.hyphaInfoButton:before {
+    content: 'i';
+}
```
matthijskooijman: There is no newline at the end of this file. Git prefers to have each line terminated, which helps when generating diffs. Ok to leave like this, but perhaps your editor can be configured to insert these newlines?

system/core/pages.php
```diff
@@ -535,3 +540,14 @@ function image_resize($src, $dst, $width, $height) {
 		}
 		return true;
 	}
+
+	function hypha_searchHelp(RequestContext $O_O, $subject, $lang = 'en') {
+		$options = [$lang, $O_O->getContentLanguage(), $O_O->getInterfaceLanguage()];
```
matthijskooijman: Should this not prefer interfacelanguage over contentlanguage?

system/core/pages.php
```diff
@@ -95,8 +95,10 @@ function newPage() {
 		html = '<table class="section"><tr><th colspan="2"><?=__('create-new-page').'<br/>'.__('instruction-new-page')?></td><tr>';
 		// TODO [LRM]: find better way to set default new page type.
 		html+= '<tr><th><?=__('type')?></th><td><select id="newPageType" name="newPageType">' + '<?php foreach($types as $type => $datatypeName) echo '<option value="'.$type.'"'.($type=='textpage' ? 'selected="selected"' : '').'>'.$datatypeName.'</option>'; ?>' + '</select></td></tr>';
-		html+= '<tr><th><?=__('pagename')?></th><td><input type="text" id="newPagename" value="<?=$pagename?>" onblur="validatePagename(this);" onkeyup="validatePagename(this); document.getElementById(\'newPageSubmit\').disabled = this.value ? false : true;"/></td></tr>';
-		html+= '<tr><td></td><td><input type="checkbox" id="newPagePrivate" name="newPagePrivate"/> <?=__('private-page')?></td></tr>';
+		var infoPageName = '<?=str_replace('\'', '\\\'', makeInfoButton('help-page-name'));?>';
```
matthijskooijman: This str_replace here probably does not catch all cases (what about backslashes?). It seems that the recommended and proper way to escape a string for javascript from PHP is to use `json_encode()` on it (which I think also generates the surrounding quotes). See https://stackoverflow.com/a/169035/740048

system/languages/nl.php
```diff
@@ -336,6 +336,13 @@
 		"art-start-review" => "start beoordeling",
 		"art-approve(s)" => "beoordeling(en)",
 		"art-unresolved-block(s)" => "onopgeloste breekpunt(en)",
+
+		// help_icon_texts
+		"help-private-page" => "een privepagina is alleen zichtbaar voor degenen die zijn ingelogd",
+		"help-version-management" => "hier kunt u terug naar een vorige versie van de pagina. Die keuze kunt u definitief maken door te klikken op Terugdraaien",
+		"help-version-revert" => "hier bevestigt u de keuze om de actuele pagina te vervangen voor deze. Wilt u dat niet? Kies dan in de versielijst weer de bovenste pagina",
+		"help-blocking-comment" => "als je een artikel blokkeert ben jij de enige die die blokkade weer weg kan nemen, bijvoorbeeld als de auteur voldoende wijzigingen heeft gemaakt",
```
matthijskooijman: Should this explain what a block does as well, in addition to how to lift the block?

system/languages/nl.php
```diff
@@ -336,6 +336,13 @@
...
+		"help-page-name" => "de paginanaam wordt gebruikt in de URL van de pagina",
```
matthijskooijman: Should this mention that the name is used as the title above the page as well?

laurensmartina:
> Note that there is still one unresolved comment (about urlencoding) from my first review. I marked all others as resolved, so it should easy to spot above.

@matthijskooijman I don't see an unresolved comment (about urlencoding). Could you please help me find it?

matthijskooijman:
Here: https://github.com/PlanBCode/hypha/pull/178#discussion_r247324354

It's hidden above, there's a block "8 hidden conversations..." in between the resolved comments.

system/languages/nl.php
dianawi: yes, I will change that

system/languages/nl.php
dianawi: I wasn't sure if we wanted to change the way that works, but for now I will change it.

system/help/help.js
laurensmartina: Added encoding on the URI.

laurensmartina:
Found #178 and pushed an update.

system/help/help.js
```diff
@@ -0,0 +1,113 @@
...
+		$infoButton.trigger('closeMe');
+		return;
+	}
+
+	// Directly set the clicked class, this also prevents the spamming of the system before the result is returned.
+	$infoButton.addClass('clicked');
+
+	// Create an empty popup.
+	var $helpPopup = $('<div></div>').appendTo('body');
+x§
```
matthijskooijman: Is this one intentional? You see to have added in the last force-push?

system/core/pages.php
```diff
@@ -187,6 +187,11 @@ function loadPage(RequestContext $O_O) {
 						break;
 				}
 				break;
+			case HyphaRequest::HYPHA_SYSTEM_PAGE_HELP:
+				$subject = isset($args[0]) ? urldecode($args[0]) : 'undefined';
```
matthijskooijman: I wonder if this is the right place to urldecode? Shouldn't HyphaRequest perhaps handle that?

system/core/pages.php
```diff
@@ -535,3 +540,14 @@ function image_resize($src, $dst, $width, $height) {
...
+
+	function hypha_searchHelp(RequestContext $O_O, $subject, $lang = 'en') {
+		$options = [$lang, $O_O->getInterfaceLanguage(), $O_O->getContentLanguage()];
+		foreach ($options as $lang) {
+			$dict = $O_O->getDictionaryByLanguage($lang);
+			if (null !== $dict && array_key_exists($subject, $dict)) {
+				return nl2br($dict[$subject]);
+			}
+		}
+		return 'Subject: "' . $subject . '" not found';
```
matthijskooijman: This needs `htmlspecialchars` to prevent XSS, I think.

system/core/various.php
```diff
@@ -110,3 +110,15 @@ function makeButton($label, $action, $id = '') {
...
+	/*
+		Function: makeInfoButton
+		Generate span with an onClick to toggle an info popup with the given tag
+
+		Parameters:
+		$tag - tag for which to get the information.
+	*/
+	function makeInfoButton($tag) {
+		global $O_O;
+		return '<span onclick="toggleInfoPopup(event, \''.$tag.'\', \'' . $O_O->getInterfaceLanguage() . '\', this)" class="hyphaInfoButton"></span>';
```
matthijskooijman: I think it would be good to escape `$tag` and the language here as well, probably using `json_encode`.

system/help/help.js
laurensmartina: Oops. :-)

system/core/pages.php
matthijskooijman: I created #203 for a proper solution (which requires more thought and testing), for now let's leave it here.

system/core/various.php
matthijskooijman: Fixed.

system/core/pages.php
matthijskooijman: Fixed.

PR #183, opened 2019-01-13
Implemented article status update message

giplt:


system/languages/en.php
```diff
@@ -285,20 +285,20 @@
 		"art-unsupported-status-change" => "this status change is not supported",
 		"art-no-approves-yet" => "there are no approves yet",
 		"art-successfully-updated" => "successfully updated",
+		"art-comment-received" => "Your comment has been received. You will receive a mail for confirmation. Click on the link in this mail to publish your comment. ",
```
matthijskooijman: This string has a trailing space.

matthijskooijman:
I did a bit more testing of this and noted that when you post a review comment, a notification is sent without a commenter name (e.g. "Test: New comment by  to 'Testarticle2'"). Turns out that the notification code assumed that a comment has a 'name' attribute, but review comments only have a 'user' attribute with a userid that must be looked up.

I made a stab at a fix, here (just a branch, no PR yet): https://github.com/PlanBCode/hypha/commits/comment-mail-wip

However, reading into the code more, it actually seemed a bit weird that this notification was sent at all. Currently, adding a new comment (review or visitor) both triggers an `EVENT_PUBLIC_COMMENT` event, which calls `onPublicComment()` which either sends a confirmation mail (when posted by a visitor), or calls `sendNewCommentMessage()` (when posted by a user).

I wonder if all this is how it was intended? `EVENT_PUBLIC_COMMENT` suggests to me "visitor comment", but apparently it is called for review comments too. This does not actually seem necessary (since review comments are always posted by users and never need confirmation), but that might change in the future? In any case, the current code (with this PR merged) will send out a (currently broken, as above) notification for *every* review comment, and a second notification mail for new blocking discussions (so you get two mails for a single comment in that case).

@giplt, did you intend to send notifications for all review comments? Or did you want to keep the old behaviour of sending a notification for new blocking discussions only and add notifications for all visitor comments only?

matthijskooijman:
The issue noted above is fixed by #192.

I noticed another issue, see https://github.com/PlanBCode/hypha/issues/99#issuecomment-455312177

dianawi:
I just left a comment on test but do not get an email.

giplt:
Apparently I didn't fully grasp the stream of events that is implemented in the article module, which seems more convoluted than I imagined. Actually I'm not so sure now where and how to implement this. Feels like the more fundamental discussion on a preferred programming paradigm needs to take place first. How do we want the datatype modules to be organised: structured, data/event-driven, other, mixed?

PR #195, opened 2019-01-19
Help icon 2

dianawi:
adds new help icons and improves previous texts.


matthijskooijman:
It seems this PR is based on a slightly older version of #178, so I think it should be updated (or perhaps these commits should just be added to #178 instead). The first commit in this PR modifies help texts introduced by #178, so it would probably be better to squash that one with the relevant commit from #178. The second commit adds a few new ones, but also further modifies the existing ones, so it could be split, but perhaps also just squashed into a single "add help icon texts" commit.

PR #197, opened 2019-01-24
More comment fixes for peer reviewed article

matthijskooijman:
This makes a few more fixes to the handling of article comments, and fixes some incorrect previous fixes. With this applied, master should be free of regressions again.


matthijskooijman:
I added one more commit adding some translations that were forgotten in a recent commit, with that I'll merge this.

Issue #199, opened 2019-01-28
Comment structure


tammoterhark:
See image. Discussion needed. The first that comes to mind is that the difference between a comment on the article and a comment on a comment is too little.

![de stadsbron test 3](https://user-images.githubusercontent.com/6483317/51836821-d3fca480-2301-11e9-964a-cd6bcce51979.png)

Possible directions for solution:
- Comment on article should be called: "Comment on the article" (same as now)
- Comment on the previous comment could be called: "Add to dialogue" or something similar, to indicate that this is within a conversation.
- a comment on the article has another  status than a comment on the comment (status = class?)
- styling should better indicate difference between the two, html to be designed
- will we allow threaded comments? How many levels?

Furthermore
- a comment or an answers by the original author of the article could have another class, to indicate that the author "speaks"

matthijskooijman:
I believe that #149 and and #177 by Bert already covers a small part of this. I'll see what part of that we can reuse from that and what needs to be different.

matthijskooijman:
> will we allow threaded comments? How many levels?

There is only a first comment in each dissussions and replies in the discussion, there is no further nesting (and I think there should not be, otherwise things become a lot more complicated).

> a comment on the article has another status than a comment on the comment (status = class?)

I believe this is already possible, the first comment has a "first-comment" CSS class.


In #149, Bert improved the rendering of closed discussions, which seems good to discuss here too. He added a label in the top right, and also used the (already existing) `closed` CSS class to render closed discussions in a different background color.

Here's what's present now:

![image](https://user-images.githubusercontent.com/194491/52088618-ac7d3480-25ac-11e9-965a-5052082b1b85.png)

@tammoterhark, how does that look?

matthijskooijman:
@tammoterhark, I think most of this is now implemented, except the author-speaks CSS class. Could you double-check and create separate new issues for anything that is yet to be implemented?

matthijskooijman:
Since most of this is implemented, and the remaining "author speaks" is not quite trivial, I'm closing this. If the wish for this feature comes up again, we can create a new issue.

PR #205, opened 2019-01-31
Article buttons at end

matthijskooijman:
This adds some buttons at the end of the article view and edit pages. It also refactors some code.

These commits are copied from #149 with some minor improvements and a rebase. Not all commits are included, some will be used later for #199 instead.

### DIFF
--- a/data/themes/default/hypha.html
+++ b/data/themes/default/hypha.html
@@ -15,7 +15,7 @@
       <div id="trail" />
       <div id="log" />
     </div>
-    <div class="container pageContainer">      
+    <div class="container pageContainer">
       <div class="flex between reverse">
         <!-- Example of a sidebar. There is CSS in place to show this
              only on the homepage (based on the id).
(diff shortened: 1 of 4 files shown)


matthijskooijman:
In particular, this adds an "approve link" at the bottom of an article view page:

image

And it adds ok/cancel buttons to the end of the article edit interface:

image


@all, how does this look (code, HTML and UI, so everyone can chip in a review :-p)

tammoterhark:
Well, it works. Why only for article?

From a editor standpoint: when defining the page and having written a lot, having the buttons at the bottom makes more sense than at the top.

Also when reviewing the page, I always rather see that the editors go all the way down, just to be sure that they have had the chance to see the whole page.

matthijskooijman:
> From a editor standpoint: when defining the page and having written a lot, having the buttons at the bottom makes more sense than at the top.

That's exactly why it's only added for article right now: all other edit interfaces are fairly short, so you usually have the top buttons in view already (and having the buttons in view twice is probably a bit of overkill). I'm happy to duplicate the buttons in other places as well, though, if we think that makes sense?

matthijskooijman:
> Also when reviewing the page, I always rather see that the editors go all the way down, just to be sure that they have had the chance to see the whole page.

@tammoterhark  that makes sense. I've just added a commit to removes the approve button from the top of the page and only shows it  at the bottom.

PR #206, opened 2019-01-31
No review in draft

dianawi:


matthijskooijman:
The commit looks good, but the PR also has some additional commits from an older version of Lauren's commit. Could you rebase on top of the latest master (e.g. `origin/master` after fetching) and make sure that all but your one commit are removed?

I also thought you were disabling the form for new review comments when the status was published (or probably, when the status is *not* under review), but I don't see that here?

matthijskooijman:
Laurens has rebased this and removed the extra commits, so looks good to merge now :-D

matthijskooijman:
Note that this question has not been answered yet, but since the PR as it is now is already useful (and since leaving review comments after an article was published does not sound like a problem per se), we've let this question block this PR. If anything should be changed in this regard, a new PR can be opened.

> I also thought you were disabling the form for new review comments when the status was published (or probably, when the status is not under review), but I don't see that here?

PR #208, opened 2019-02-02
Added styling for texarea

### LINKED ISSUE #207: Textarea for CSS and for hypha.html show wrong font
The current fonts shown here are not monospaced, while for code like text areas they should be.

tammoterhark:
In order to produce better presentation in text areas.

This fixes #207

### DIFF
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -694,6 +694,11 @@ th{
  flex-direction: row-reverse;
 }
 
+textarea{
+	font-family: monospace;
+	font-size: initial;
+}
+
 /* Festival */
 .festivalTimetable, .festivalTimetable a {
 	color: white;


matthijskooijman:
This looks like a good fix. However, two remarks:
 - I'm not such a fan of the Courier font. It quickly springs to mind when thinking of monospaced fonts, but it is also a serif font, which isn't all that suitable for displaying HTML code, I believe. I would suggest just specifying `font-family: monospace` and let the browser figure out what monospace font to use. I checked Firefox and Chrome, both will give you the default textarea font in this case (I also tried `font-familiy: initial` to force back to the default, but it seems that results in the normal default font, not the default font for textareas).
 - In addition to the font, normalize also changes the text size to something a lot smaller (and IMHO less usable) than the default. I would suggest reverting back to the default font size (which seems to work with `font-size: initial`).

For comparison, here's what it looks like with the current PR from Tammo:

image

With my above suggestions, it looks like this:

image

@tammoterhark, let me know what you think of my suggestions. If you think they're ok, I can update and then merge this PR with those changes (I already have them committed locally).

tammoterhark:
Completely agree. Thanks for merging.

matthijskooijman:
I've updated this PR, expanded the commit message a bit and rebased on top of the latest master, should be ready to merge now.

PR #214, opened 2019-02-16
peer_reviewed_article: display public discussion count

laurensmartina:
The number of discussions is displayed above the article title.


tammoterhark:
For now I would opt to show all comments, not just the discussions, as a very lively discussion may stay just one discussion, but may draw many comments. 

On the page (and later the front page) a growing number of comments may draw readers. BOTOH: we seek quality, not quantity as such. Should we let the user group (De Stadsbron redactie) decide?

matthijskooijman:
For now, it seems best to count the number of comments, rather than the number of discussions. This needs some changes in the code.

matthijskooijman:
Looks good!

PR #215, opened 2019-02-16
Changed login form html structure to div elements

laurensmartina:
The login form was structured in a table which is not easy to style.
In order to be more flexible with the way the login form can be styled
the html structure needed to be updated to div elements.

### DIFF
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -318,9 +318,7 @@ a.nopagelink {
  position: absolute;
  top: 50px;
  background: #FFFFE0;
- left: 10%;
  margin-right: auto;
- border: 1px solid darkgrey;
  z-index: 1;
 }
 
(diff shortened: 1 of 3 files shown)


tammoterhark:
Not sure if this is within the scope of this PR, but ideaslly there should be no styling in the HTML. Currently there is in the HTML:

`<div id="popup" style="left: 0px; top: 25px; visibility: visible;">`

Please also take out the style statements.

matthijskooijman:
@tammoterhark, you seem to have moved button CSS to another place in the CSS file. Initially, I thought it was accidental, or at least unrelated to this change, but later I noticed that you're overriding the .button handling for the login form, so I suppose that the button CSS needs to be above the login form, hence you moved it. In this case, it would have been better to mention this in the commit message, or even put this in a separate commit, so it is easier to say something like "This only moves CSS, without any modifications".

However, I actually wonder if we should be reusing the same `button` class here. In other places, there is an `input` element with a `button` class, but here we're adding a `div` with `button` class *around* a button, which is actually a different kind of thing. So it's probably best to make sure that the same rules do not apply to both (either by doing `input.button` and `div.button`, or replacing the new `button` class with `button_wrapper` perhaps?). Though I wonder, do we really need that wrapper at all? You now use it to put the buttons side by side with `inline-block` and add some padding in between, but I'd think you could just make the input element `inline` (isn't that even the default) and use margin?

I also noticed some changes to the `#popup` CSS, but as we discussed this afternoon, I think we should move this into its own PR later.

laurensmartina:
> * I wonder if this changes makes sense in `hypha.php`, since that does not include the main CSS but instead has a bit of minimal inline CSS that now no longer applies. We could change this in `hypha.php` for consistency with the main code, but that would break consistency with the rest of `hypha.php` (that still uses tables).

Removed changes in `hypha.php`.

data/themes/default/hypha.css
```diff
@@ -350,6 +347,44 @@ button: last-of-type{
...
+}
+
+.login-wrapper button,
+.login-wrapper [type="submit"],
+.login-wrapper [type="button"],
+.login-wrapper [type="reset"],
+.login-wrapper [type="cancel"]{
+	-webkit-appearance: none;
+   -moz-appearance: none;
+   appearance: none;
```
matthijskooijman: There are mixed spaces/tabs here.

matthijskooijman: Also, all this new CSS uses tabs or 3 spaces for indentation, where the rest of the file uses single spaces.

laurensmartina: fixed

PR #216, opened 2019-02-16
Adds class names to elements on settings page

laurensmartina:
To style certain elements on the settings page, some class are added.

### DIFF
--- a/system/core/various.php
+++ b/system/core/various.php
@@ -95,8 +95,9 @@
 		$action - javascript action
 		$id - optional element id
 	*/
-	function makeButton($label, $action, $id = '') {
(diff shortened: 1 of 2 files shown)


tammoterhark:
I guess you are right @matthijskooijman 

@laurensmartina could you please change the html of the two buttons below the personal settings from:
`class = "button right"` tp 
`class= "button edit"` and
`class= "button stop"` respectively?

laurensmartina:
> @laurensmartina could you please change the html of the two buttons below the personal settings from:
> `class = "button right"` to
> `class= "button edit"` and
> `class= "button stop"` respectively?
done! :-)

matthijskooijman:
We just made one more change: Rather than "stop" we used "quit" for the CSS class, since that is more clear and matches the string id. Laurens is now looking at improving the button text itself (stop in Dutch and unsubscribe in English isn't entirely clear right now).

PR #218, opened 2019-02-23
Adds an explanation to the quit page

laurensmartina:
At the quit page, it was unclear what the user could expect.
This fixes that problem.


matthijskooijman:
The commit also changes the button text, @laurensmartina, perhaps you could add that in the commit message?

@all, what do you think of this change/wording? Here's how it looks with this PR:

image
image
image
image


Personally, I wouldn't mind removing the "Screw you guys..." text, just leave the textbox empty by default?

laurensmartina:
I think it's fine for now.. It can be further improved if needed.

dianawi:
I think it is good, and yes, let's take out the screw you guys text.

matthijskooijman:
I've removed the "Screw you..." wording, and rebased. Ready to merge now.

PR #219, opened 2019-02-23
Remove cols and rows attributes from comment textareas

### LINKED ISSUE #217: Inline style in textarea hinders styling
Currently a textarea for leaving comment is shown inline as:

`<textarea name="discussion_comment/id5c712d8646f15" id="discussion_comment/id5c712d8646f15" cols="40" rows="9"></textarea>`

This inline styling hinders styling. I would propose to change this element to:

For review comment:
`<textarea name="discussion_comment/id5c712d8646f15" id="discussion_comment/id5c712d8646f15" class="review-comment body"></textarea>` 

For reader comment:
`<textarea name="discussion_comment/id5c712d8646f15" id="discussion_comment/id5c712d8646f15" class="reader-comment body"></textarea>`

matthijskooijman:
This fixes #217

### DIFF
--- a/system/datatypes/peer_reviewed_article.php
+++ b/system/datatypes/peer_reviewed_article.php
@@ -1167,7 +1167,7 @@ EOF;
 		$html = <<<EOF
 			<div class="new-comment $type">
 			<div>
-				<strong><label for="$commentFieldName"> $comment $infoCommentRules</label></strong><textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
+				<strong><label for="$commentFieldName"> $comment $infoCommentRules</label></strong><textarea name="$commentFieldName" id="$commentFieldName"></textarea>
 			</div>
 EOF;
 		if ($new && self::FIELD_NAME_DISCUSSION_REVIEW_CONTAINER === $type && !in_array($this->getStatus(), [self::STATUS_APPROVED, self::STATUS_PUBLISHED])) {
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -851,3 +851,8 @@ body.type_peer_reviewed_article .input-wrapper.field_type_editor:not(.field_name
 .hyphaInfoButton:before {
     content: 'i';
 }
+
+.new-comment textarea{
+	width: 100%;
+	height: 100px;
+}


matthijskooijman:
@tammo, this removes the cols and rows attributes as suggested in #217. Could you have a look at adding the proper CSS to make the textareas look ok again in the default CSS?

Also, there are some more textareas with rows and cols in the settings page. I was planning to also remove these here, but I think there are not sufficient CSS classes there to be able to style them properly, so I left them out for now.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1142,7 +1142,7 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
 		$html = <<<EOF
 			<div class="new-comment $type">
 			<div>
-				<strong><label for="$commentFieldName"> $comment </label></strong><textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
+				<strong><label for="$commentFieldName"> $comment </label></strong><textarea name="$commentFieldName" id="$commentFieldName"></textarea>
 			</div>
```
tammoterhark: The id that is produced here cannot be handled with css. 

What is produced:

<div>
--
```
 <strong><label for="discussion_comment/new_public"> reageer op het artikel </label></strong><textarea name="discussion_comment/new_public" id="discussion_comment/new_public"></textarea>
 </div>
```
In my opinion the id="discussion_comment new_public" would be better, since the slash produces an error in my css editor.

tammoterhark: Decided to delete the ID, since it was not needed.

matthijskooijman:
I squashed the CSS changes into the first commit (so it now removes rows/cols and replaces that with CSS in one go).

I had another look at the second commit removing the ID tag, but it seems it is required after all. Not for CSS matching, but to match the `label`'s `for` attribute. As for the contents of the id attribute, according to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) whitespace is disallowed, but (in HTML5 at least) anything else is ok. So I think the current code is actually fine?

I would propose dropping the second commit and merging the first (after one more test).

matthijskooijman:
I rebased, dropped the second commit (which removed the id attribute), so this can be merged.

PR #221, opened 2019-03-02
Adapted the spot where nr-of-comments is placed via hypha.html

tammoterhark:
Instead of just placing the number of comments at a relatively abritrary position, we created a div with an ID="nr-of-comments" that can be placed via hypha.html.

### DIFF
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -22,7 +22,6 @@
  --menulinkhoverback: #aaa;
  --menulinkactiveback: #ccc;
  --sidebarbackgroundcolor: #ddddff;
- 
 }
 
 /* dosis-regular - latin */
@@ -841,4 +840,8 @@ body.type_peer_reviewed_article .input-wrapper.field_type_editor:not(.field_name
 }
 .hyphaInfoButton:before {
     content: 'i';
+}
+
+#nr-of-comments p{
+	margin: 0;
 }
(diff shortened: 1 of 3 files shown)


matthijskooijman:
This sounds like a useful feature to me, though I'm not so sure I like the way it works now. In particular, now *every* generated page has a `nr-of-comments` div, most of them empty and not applicable to the generated page... I'm afraid that putting this kind of thing in hypha.xml is not really the right approach here (it also doesn't scale very well), but I'm not sure if there's anything better.

What do others think?

Looking at the specific problem you want to solve: Where would you like to put the nr of comments? Perhaps we can just move it in the code, rather than make it configurable?

As for the actual commits: Note that the hypha.xml file is completely changed (every line has a change), presumably because you've switched from dos to unix line endings or vice versa. Please try to avoid that, and make sure to review your commits after you make them (e.g. look at the diff) to spot unexpected changes like these.

tammoterhark:
We found out that when you use the internal editor for changing either hypha.html of hypha.css, the whole file is seen as "new". Editing only externally will overcome this problem. But actually it should not happen, of course...

tammoterhark:
OK, let's not use this PR, since it is too specific, only applicable for articles.

This may open a discussion for the need for specific templates for page types, something like the following:

All pages use hypha.html, except when there is a more specific template, like peer-reviewed-article.html or textpage.html. 

Food for thought for later.

Issue #107, opened 2018-11-26
Explain page types


tammoterhark:
When wanting to define a new page, the content type name may be not enough to make a right decision about what to use.

In order to make a good choice, please change the current dropdown for choosing a new piece of content to a page where each title can be followed by a localised explanation of what the content type is intended for.

PR #223, opened 2019-03-02
Adds help icon for choosing new page type

dianawi:


system/languages/en.php
```diff
@@ -358,6 +358,7 @@
 
 		// help_icon_texts
 		"help-private-page" => "A private page is only visible to users who are logged in.",
+		"help-page-type" => "choose your pagetype first. \nA festival page will give you the tools to organise a festival and show the line-up to the visitor. \nA web page is a regular weg page. \nA mailinglist creates a list for emailaddresses and a tool to send mailings. \nA peer reviewed article creates a page that remains private until 3 people have approved it, \nthen it becomes public with a comments possibility. \nAweb page creates a regular web page.",
```
matthijskooijman: I think each line of this helptext should start with a capital letter? And I think there is no point in having spaces before each `\n`?

I also don't really like hardcoding this list of possible page types, since we (or users) might add new datatypes which are then not reflected here. However, to properly solve this, datatypes must be able to provide their own helptext fragment somehow. But that is complicated, so until we have that, this is probably a fine approach (but we should create an issue to track this once this is merged).

matthijskooijman:
I think this fixes #107, so it might be good to note that in the commit message (using the github auto-close syntax).

laurensmartina:
Fixed all, please review again.

matthijskooijman:
Looks good, I think this can be merged after testing.

matthijskooijman:
I made one more textual change (the help text said 'artikel met peer reviewed', while the dropdown it explains says 'artikel met review') and rebased this on top of master. Merging now.

PR #224, opened 2019-03-02
Sloppy language

dianawi:


system/languages/en.php
```diff
@@ -257,6 +260,7 @@
 		"ml-successfully-unsubscribed" => "you have successfully unsubscribed",
 		"ml-successfully-created" => "successfully created",
 		"ml-successfully-updated" => "successfully updated",
+		"ml-comment-please-confirm" => "Thank you for your comment. Please confirm your identity by clicking the confirmation links that we have sent you by email.",
```
matthijskooijman: I can't find any reference to this string in the code?

matthijskooijman: Ah, seems this was added in #227 (both the string definition and the usage). I think this should be dropped from this commit.

matthijskooijman: Scratch that, it seems #227 only adds the strings, but does not use them anywhere...

tammoterhark: Looks to me like `ml` refers to mailinglists. Can they have comments?

matthijskooijman:
Looks good at first glance. I think this can be merged after testing.

matthijskooijman:
Tested, works. I rebased on top of master, merging now.

PR #225, opened 2019-03-02
Further improve theming support

laurensmartina:
Moved help.js to assets directory.
Remove option to edit the default theme.
Themes can be copied through the web interface.
Adds option to preview a theme.


system/datatypes/settings.php
```diff
@@ -304,16 +306,29 @@ function editTheme() {
 			if (isAdmin()) {
 				ob_start();
 				echo '<select name="editTheme" id="editTheme">';
-				foreach (scandir('data/themes/') as $theme) {
-					if ('.' !== $theme[0]) {
-						echo '<option' . (Hypha::$data->theme === $theme ? ' selected' : '') . '>' . htmlspecialchars($theme) . '</option>';
-					}
+				$themes = $this->getThemes();
+				foreach ($themes as $theme) {
+					echo '<option' . (Hypha::$data->theme === $theme ? ' selected' : '') . '>' . htmlspecialchars($theme) . '</option>';
```
matthijskooijman: This is a good change, but looks like it should be in a separate commit.

system/datatypes/settings.php
```diff
@@ -327,6 +342,109 @@ function saveTheme($argument) {
...
+			// Strip out any url-encoded stuff
+			$dir = urldecode($dir);
+			// Replace non-AlNum characters with space
+			$dir = preg_replace('/[^A-Za-z0-9]/', ' ', $dir);
+			// Replace Multiple spaces with single space
+			$dir = preg_replace('/ +/', ' ', $dir);
+			// Trim the string of leading/trailing space
+			$dir = trim($dir);
+			// Replace spaces with dashes
+			$dir = preg_replace('/ /', '-', $dir);
```
matthijskooijman: Why do the first three things? There is no reason for any of these encodings to be present AFAICS, so they could only be there when a user types them manually. In that case, the next regex (non-alnum) will strip away such encodings anyway to make things harmless. In fact, I think *only* the alnum regex could actually be sufficient already (it would be safe), though the operations after that could make a string with non alnum stuff nicer to read (but it might still be unexpected for a user that his carefully typed string is chopped up). Perhaps it is nicer to do this checking client-side already (either flag invalid values or prevent typing invalid chars in the first place). I suspect that HTML5 might have something convenient for this? Of course, a server side check/replace is still needed, but that can then just be the alnum regex above, no need to generate something pretty, only safe.

We could consider allowing both - and _ in names as well? . is also obvious, but opens the door to path traversal (though maybe not without /).

system/datatypes/settings.php
```diff
@@ -268,10 +268,19 @@ function saveHyphaSettings() {
...
 			if (isAdmin()) {
-				$this->html->writeToElement('pagename', __('edit-html'));
 				$this->html->writeToElement('pageCommands', makeButton(__('cancel'), makeAction('settings', '', '')));
-				$this->html->writeToElement('pageCommands', makeButton(__('save'), makeAction('settings', 'settingsSaveMarkup', '')));
-				ob_start();
+				if (Hypha::$data->theme === 'default') {
+					global $hyphaUrl;
+					$this->html->writeToElement('pagename', __('view-html-of-theme', ['theme' => Hypha::$data->theme]));
+					$this->html->writeToElement('main', __('cannot-edit-default-theme-explanation', ['link' => $hyphaUrl.'settings/theme']));
+					ob_start();
```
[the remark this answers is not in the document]
laurensmartina: I've moved it, although this is also fine for me.

system/datatypes/settings.php
```diff
@@ -327,6 +342,109 @@ function saveTheme($argument) {
...
+			if (!isset($_POST['srcTheme']) || '' == $_POST['srcTheme']) {
+				$errors[] = __('source-theme-name-required');
+			} elseif (!in_array($_POST['srcTheme'], $themes)) {
+				$errors[] = __('source-theme-name-not-found');
+			}
+
+			// return the errors if there are any
+			if (empty($errors)) {
+				// create an anonymous function to cursively copy the theme directory
+				$dirCopy = function ($src, $dst) use (&$dirCopy) {
```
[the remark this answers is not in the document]
laurensmartina: Agreed, we should move this, I'll add a todo.

system/datatypes/settings.php
laurensmartina: This is a copy-paste from https://stackoverflow.com/a/7128879, I didn't want to spend too much time digging into this.

system/datatypes/settings.php
laurensmartina: I introduced the function in this commit, might as well use it here.

system/datatypes/settings.php
```diff
@@ -290,6 +293,7 @@ function editMarkup() {
 	</tr>
 </table>
 <?php
+				}
```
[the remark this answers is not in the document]
laurensmartina: Fixed

laurensmartina:
> I did a more thorough review just now, adding a few more comments.
> 
> The UI could indeed use some improvements, though I don't have any suggestions right now. I do wonder if offering both 'copy' and 'copy and activate' is really useful and clear to users?

This suggestion was made by @tammoterhark, I don't have an opinion.

laurensmartina:
> Hm, seems I missed the preview commit in my last review. I had a look at it now, and I wonder if the theme stuff needs some refactoring. In particular, I think the global `hypha_getTheme()` (and the related functions introduced now perhaps) should be moved into `RequestContext`? Also, most of the file instances in the `Hypha` class should probably be moved to `RequestContext` as well.

We should talk about this further, I don't think this would it within the scope of RequestContext.

system/datatypes/settings.php
laurensmartina: Solved in an other way.

PR #227, opened 2019-03-09
Comment unanonymous 2

dianawi:


system/datatypes/peer_reviewed_article.php
```diff
@@ -1182,10 +1182,13 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
...
 			$email = __('art-email');
 			$emailFieldName = self::FIELD_NAME_DISCUSSION_COMMENTER_EMAIL;
+			$unanonymous =  __('art-comment-unanonymous');
 			$html .= <<<EOF
 			<div class="section" style="padding:5px; margin-bottom:5px; position:relative;">
-				<strong><label for="$nameFieldName"> $name </label></strong> <input type="text" id="$nameFieldName" name="$nameFieldName" />
+				<strong><label for="$nameFieldName"> $name $unanonymous </label>  </strong><input type="text" id="$nameFieldName" name="$nameFieldName" />
 			</div>
+EOF;
+			$html .= <<<EOF
```
matthijskooijman: Why break the EOF-block here? IIRC this change is not needed?

matthijskooijman: Why break the EOF-block here? AFAICS this change is not needed?

system/datatypes/peer_reviewed_article.php
```diff
@@ -1182,10 +1182,13 @@ private function getCommentFormHtml($type, DocPage $discussion = null) {
 			$nameFieldName = self::FIELD_NAME_DISCUSSION_COMMENTER_NAME;
 			$email = __('art-email');
 			$emailFieldName = self::FIELD_NAME_DISCUSSION_COMMENTER_EMAIL;
+			$unanonymous =  __('art-comment-unanonymous');
 			$html .= <<<EOF
 			<div class="section" style="padding:5px; margin-bottom:5px; position:relative;">
-				<strong><label for="$nameFieldName"> $name </label></strong> <input type="text" id="$nameFieldName" name="$nameFieldName" />
+				<strong><label for="$nameFieldName"> $name $unanonymous </label>  </strong><input type="text" id="$nameFieldName" name="$nameFieldName" />
```
matthijskooijman: I wonder if this unanonymous remark should be in the `<label>` tag, or perhaps outside of it? I wonder if it is conceptually part of the label or not? Or should this perhaps be an info-icon?

system/languages/en.php
```diff
@@ -257,6 +257,7 @@
 		"ml-successfully-unsubscribed" => "you have successfully unsubscribed",
 		"ml-successfully-created" => "successfully created",
 		"ml-successfully-updated" => "successfully updated",
+		"ml-comment-please-confirm" => "Thank you for your comment. Please confirm your identity by clicking the confirmation links that we have sent you by email.",
```
matthijskooijman: This commit only seems to add this string, but never uses it?

laurensmartina:
Fixed all, please review again.

matthijskooijman:
Works, rebased and merging.

PR #229, opened 2019-03-09
Implementation of tagging system

giplt:
Adds a tag index to a datatype similar to adding a page index. Still lacking in this implementation:
- add styling for default theme, @tammoterhark: can you take a look at this?
- implement error notification if new tag already exists in the tagList
- implement tag management page: edit, translate, remove or merge tags. Can be done (for the time being) by editing hypha.xml.
- implement tagcontainer pagetype.

### DIFF
--- a/system/core/events.php
+++ b/system/core/events.php
@@ -48,8 +48,9 @@
 		cmd - php function to execute, e.g. 'save'
 		arg - argument to pass to the cmd function
 		form - the form tag to use for posting (if cmd or arg are given)
+		callback - optional callback function for making ajax calls
 	*/
-	function hypha(url, cmd, arg, form) {
+	function hypha(url, cmd, arg, form, callback = null) {
 		// update url
 		url = url.replace(/\s\//g, '/').replace(/\s$/g, '').replace(/\s/g, '_');
 
@@ -76,6 +77,11 @@
 			$form.append($arg);
 		}
 		$arg.val(arg);
+
+		if (callback) {
+			ajax(url, callback, true);
+			return;
+		}
 
 		// When there is a field with name "submit",
 		// $form.submit (and $form[0].submit) will be a
--- a/system/core/tags.php
+++ b/system/core/tags.php
@@ -1,0 +1,195 @@
+<?php
+	// TODO:
+	// - implement error notification if new tag exists
+	// - implement tag management page: edit, translate, remove or merge tags
+	// - implement tagcontainer pagetype
+
+	function tagList(HyphaDomElement $pageListNode, $lang) {
+		/** @var HyphaDomElement $hyphaXml */
+		global $hyphaXml;
+
+		$selectedTagIds = [];
+
+		/** @var HyphaDomElement $tagEl */
+		// Build list of active tags for given page
+		foreach ($pageListNode->findXPath('tag') as $tagEl) $selectedTagIds[] = $tagEl->getAttribute('id');
+
+		$selectedTags = [];
+		$deselectedTags = [];
+
+		// Divide list of all tags stored in hypha.xml in a selected tag list and a deselected tag list
+		foreach ($hyphaXml->findXPath('hypha/tagList/tag/language[@id=' . xpath_encode($lang) . ']') as $tagLang) {
+			/** @var HyphaDomElement $tagLang */
+			$id = $tagLang->parent()->getId();
+			$tag = ['label' => $tagLang->getAttribute('label'), 'description' => $tagLang->nodeValue];
+			if (in_array($id, $selectedTagIds)) $selectedTags[$id] = $tag;
+			else $deselectedTags[$id] = $tag;
+		}
+
+		// Construct html output
+		$html = '';
+		if (!empty($selectedTagIds) || isUser()) {
+			$html.= '<span class="prefix">' . __('tags') . ': </span>';
+			$html.= '<div class="selectedTags">';
+			foreach ($selectedTags as $id => $tag) {
+				$html.= '<div class="tagSel_'.htmlspecialchars($id).'" class="tag">';
+				$html.= '<span title="'.htmlspecialchars($tag['description']).'">'.htmlspecialchars($tag['label']).'</span>';
+				if (isUser()) $html.= ' <span class="delete-tag" title="'.__('remove-tag').'" onclick="deselectTag('.htmlspecialchars(json_encode($id)).');">⨯</span>';
+				$html.= '</div>';
+			}
+			$html.= '</div>';
+			if ((isUser() && !empty($deselectedTags)) || isAdmin()) {
+				$html.= '<select class="remainingTags" onchange="selectTag(this, this.value);">';
+				$html.= '<option value="_add_" selected="selected" disabled="disabled">'.__('add').'...</option>';
+				foreach ($deselectedTags as $id => $tag) {
+					$html.= '<option id="tagRem_'.htmlspecialchars($id).'" value="'.htmlspecialchars($id).'">'.htmlspecialchars($tag['label']).'</option>';
+				}
+				if (isAdmin()) $html.= '<option value="">'.__('new').'...</option>';
+				$html.= '</select>';
+			}
+		}
+		return $html;
+	}
+
+	/*
+		Function: hypha_indexTags
+		returns list of tags for the given page
+
+		Parameters:
+		$pageListNode - DOMElement containing page settings
+		$lang - language for tag index
+	*/
+	function hypha_indexTags($pageListNode, $lang) {
+		global $hyphaHtml;
+		$formId = $hyphaHtml->getDefaultForm()->getId();
+
+		if (!$pageListNode || !$lang) return '';
+
+		$language = hypha_pageGetLanguage($pageListNode, $lang);
+		$url = $language->getAttribute('id').'/'.$language->getAttribute('name');
+
+		ob_start();
+?>
+<script>
+	function selectTag(select, id) {
+		if (!id) newTag(select);
+		else {
+			hypha(<?=json_encode($url)?>, 'tagSelect', id, document.getElementById(<?=json_encode($formId)?>), function(response) {document.getElementById('tagList').innerHTML = response;});
+			document.getElementById('tagList').innerHTML = 'loading...';
+		}
+	}
+	function deselectTag(id) {
+		hypha(<?=json_encode($url)?>, 'tagDeselect', id, document.getElementById(<?=json_encode($formId)?>), function(response) {document.getElementById('tagList').innerHTML = response;});
+		document.getElementById('tagList').innerHTML = 'loading...';
+	}
+	function newTag(select) {
+		// Revert back to initial value, so you can choose "new" again
+		select.options[0].selected = true;
+		var name = prompt(<?=json_encode(__('enter-tag-label'))?>, '');
+		if (name) {
+			var description = '';
+			while (description === '') {
+				// null when user clicks cancel
+				var description = prompt(<?=json_encode(__('enter-tag-description'))?>, '');
+			}
+
+			if (description) {
+				var json = JSON.stringify({name: name, description: description});
+				hypha(<?=json_encode($url)?>, 'tagNew', json, document.getElementById(<?=json_encode($formId)?>), function(response) {document.getElementById('tagList').innerHTML = response;});
+				document.getElementById('tagList').innerHTML = 'loading...';
+			}
+		}
+	}
+</script>
+<?php
+		$hyphaHtml->writeScript(ob_get_clean());
+
+		return tagList($pageListNode, $lang);
+	}
+
+	registerCommandCallback('tagSelect', 'tagSelectCallback');
+	registerCommandCallback('tagDeselect', 'tagDeselectCallback');
+	registerCommandCallback('tagNew', 'tagNewCallback');
+
+	function tagSelectCallback($id) {
+		global $hyphaXml;
+		global $O_O;
+
+		$lang = $O_O->getContentLanguage();
+
+		$hyphaXml->lockAndReload();
+		$pageListNode = hypha_getPage($lang, $O_O->getRequest()->getPageName());
+		$newtag = $hyphaXml->createElement('tag');
+		$newtag->setAttribute('id', $id);
+		$pageListNode->appendChild($newtag);
+		$hyphaXml->saveAndUnlock();
+
+		echo tagList($pageListNode, $lang);
+		exit;
+	}
+
+	function tagDeselectCallback($id) {
+		global $hyphaXml;
+		global $O_O;
+		$lang = $O_O->getContentLanguage();
+
+		$hyphaXml->lockAndReload();
+		$pageListNode = hypha_getPage($lang, $O_O->getRequest()->getPageName());
+		foreach($pageListNode->getElementsByTagName('tag') as $tag) if ($tag->getAttribute('id')==$id) $pageListNode->removeChild($tag);
+		$hyphaXml->saveAndUnlock();
+
+		echo tagList($pageListNode, $lang);
+		exit;
+	}
+
+	function tagNewCallback($json) {
+		global $hyphaXml;
+		global $O_O;
+
+		$hyphaXml->lockAndReload();
+
+		$lang = $O_O->getContentLanguage();
+		$pageListNode = hypha_getPage($lang, $O_O->getRequest()->getPageName());
+
+		$json = json_decode($json, true);
+
+
+		// Check if tagList exists
+		$tagList = $hyphaXml->findXPath('hypha/tagList');
+		if ($tagList->count() === 0) {
+			$tagList = $hyphaXml->createElement('tagList');
+			$hyphaXml->documentElement->appendChild($tagList);
+		}
+
+		// Check if tag exists
+		$path = 'hypha/tagList/tag/language[@id=' . xpath_encode($lang) . ' and @label=' . xpath_encode($json['name']) . ']';
+		$tagExists = $hyphaXml->findXPath($path)->count() > 0;
+
+		// Create new tag, add to tagList and page in hypha.xml
+		if (!$tagExists) {
+			$newTagName = $json['name'];
+			$newTagDescription = $json['description'];
+
+			$newTag = $hyphaXml->createElement('tag');
+			$newTag->generateId();
+			$newTagLanguage = $hyphaXml->createElement('language', $newTagDescription);
+			$newTagLanguage->setAttribute('id', $lang);
+			$newTagLanguage->setAttribute('label', $newTagName);
+			$newTag->appendChild($newTagLanguage);
+			$tagList->appendChild($newTag);
+
+			$newTagRef = $hyphaXml->createElement('tag');
+			$newTagRef->setAttribute('id', $newTag->getId());
+			$pageListNode->appendChild($newTagRef);
+
+			$hyphaXml->saveAndUnlock();
+		} else {
+			$hyphaXml->unlock();
+		}
+
+		// TODO: Replace this with a return value that lets the
+		// caller know we want to return a response (but still
+		// allow any postprocessing or other cleanup).
+		echo tagList($pageListNode, $lang);
+		exit;
+	}
--- a/system/languages/en.php
+++ b/system/languages/en.php
@@ -201,6 +201,8 @@
 		"invalid-argument" => "invalid argument",
 		"archive" => "archive",
 		"sure-to-delete" => "Are you sure you want to delete this page",
+		"enter-tag-label" => "Enter new tag (use a short word)",
+		"enter-tag-description" => "Enter short description (required)",
 		"login-to-perform-action" => "you need to be logged on in order to perform this action",
 		"page-successfully-updated" => "Page successfully updated",
 		'cannot-edit-default-theme-explanation' => 'The default theme cannot be edited. At the <a href="[[link]]">theme selection page</a>, the default theme can be copied and activated, that copy can be edited.',
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -298,7 +298,7 @@ a.nopagelink {
  margin: 30px 0 10px;
  padding-top: 5px;
  border-top: 1px dotted #222;
- font-size: 8pt;
+ font-size: 80%;
 }
 
 #footer {
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -292,6 +292,39 @@ a.nopagelink {
     border-top: 1px dotted grey;
 }
 
+#tagList .prefix{
+    margin-left: 10px;
+    margin-bottom: 5px;
+    font-size: 80%;
+}
+
+.selectedTags {
+    display: flex;
+}
+
+.selectedTags .delete-tag {
+    color: #8f0000;
+    cursor: pointer;
+}
+
+.remainingTags {
+    padding: 0;
+    font-size: 80%;
+}
+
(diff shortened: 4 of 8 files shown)


system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+		
+		// Construct html output
+		$html = '';
+		if (!empty($activeTags) || isUser()) {
+			$html.= '<span class="prefix">' . __('tags') . ': </span>';
+			$html.= '<div id="selectedTags" style="display:inline-block;">';
+			if (!empty($selectedTags)) foreach ($selectedTags as $id => $tag) {
+				$html.= '<div id="tagSel_'.$id.'" class="tag">';
+				$html.= '<span title="'.$tag['description'].'">'.$tag['label'].'</span>';
+				if (isUser()) $html.= ' <span style="color:#8f0000; cursor:pointer;" title="'.__('remove-tag').'" onclick="deselectTag(\''.$id.'\');">⨯</span>';
```
tammoterhark: Please do not use inline styling, this can better be tackled in CSS. Better: 

<span class="delete-tag" />

That way the stylist can put here anything (I think)

system/languages/en.php
```diff
@@ -192,6 +192,8 @@
 		"missing-arguments" => "missing arguments",
 		"archive" => "archive",
 		"sure-to-delete" => "Are you sure you want to delete this page",
+		"enter-label" => "Enter label (use a short word)",
```
dianawi: I think it is clearer to use the word Tag here instead of Label. Also it is nice to make the communication with users a bit more human, so I would suggest "Please enter a label/tag" and "Enter a short description". The same goed for the Dutch translation.

system/core/events.php
```diff
@@ -53,12 +53,6 @@ function addEventHandler($html) {
 		ob_start();
 ?>
 <script>
-	/*
-		Variable: baseUrl
-		Javascript variable containing the base url from the document location object.
-	*/
-	var postProcessingList = new Array();
```
matthijskooijman: This removes the `postProcessingList` without comment. Is it unused and therefore removed? Is it replaced by something else? This should probably be a separate commit in any case?

system/core/events.php
```diff
@@ -67,15 +61,18 @@ function addEventHandler($html) {
 		url - the hypha-formatted page identifier, e.g. 'en/hompage/defaultview'. When left empty the current paged is reloaded.
 		cmd - php function to execute, e.g. 'save'
 		arg - argument to pass to the cmd function
+		callback - optional callback function for making ajax calls
 	*/
-	function hypha(url, cmd, arg) {
+	function hypha(url, cmd, arg, callback = null) {
```
matthijskooijman: Adding support for a callback here should probably be a separate commit (not a separate PR, though).

I also wonder if adding a callback here is clear enough, since now adding the callback argument in a call completely changes the behaviour of this function. I wonder if it would be better to add the cmd and arg arguments to the `ajax` function (or add a flavour of that function with those arguments)?

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+	// - add styling for default theme
+	// - implement error notification if new tag exists
+	// - implement tag management page: edit, translate, remove or merge tags
+	// - implement tagcontainer pagetype
+	
+	function tagList($pageListNode, $lang) {
+		global $hyphaXml;
+		
+		// Build list of active tags for given page
+		foreach($pageListNode->getElementsByTagName('tag') as $tag) $activeTags[] = $tag->getAttribute('id');
```
matthijskooijman: I believe getElementsByTagName returns all descendent tags, not just direct children. Is that intended here? It probably doesn't hurt, though.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+		
+		// Divide list of all tags stored in hypha.xml in a selected tag list and a deselected tag list
+		foreach($hyphaXml->getElementsByTagName('tag') as $tag) {
```
matthijskooijman: This also finds *all* tag elements, which I think is not intended here? E.g. it finds the tag definitions, but also the tag references inside the page list. In practice the latter do not have any `language` children elements, so this doesn't break, but it is at least inefficient and prone to breakage in the future.

I would suggest using Xpath, e.g. something like `$hyphaXml->findXPath("/tags/tag")` (not tested, but I think `$hyphaXml` should have a `findXPath()` method, and I'm just guessing about the XML structure that you need to match with this Xpath expression).

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+			foreach($tag->getElementsByTagName('language') as $tagLang) {
+				if ($tagLang->getAttribute('id')==$lang) {
+					if (!empty($activeTags) && in_array($tag->getAttribute('id'), $activeTags)) $selectedTags[$tag->getAttribute('id')] = array("label" => $tagLang->getAttribute('label'), "description" => $tagLang->nodeValue);
```
matthijskooijman: This seems like an overly long line to me, I'd add some line break(s).

Also, I think that if you assign an empty array to `$activeTags` at the start of the function, you won't have to do the `empty` check here.

system/core/tags.php
matthijskooijman: I wonder if it would be a good idea to let the xpath query lookup the tags by language (and perhaps id)? That might make the code simpler.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+			}
+		}
+		
+		// Construct html output
+		$html = '';
+		if (!empty($activeTags) || isUser()) {
+			$html.= '<span class="prefix">' . __('tags') . ': </span>';
+			$html.= '<div id="selectedTags" style="display:inline-block;">';
+			if (!empty($selectedTags)) foreach ($selectedTags as $id => $tag) {
+				$html.= '<div id="tagSel_'.$id.'" class="tag">';
```
matthijskooijman: This needs `htmlspecialchars` on the `$id`, and also the tag description and label below.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+		}
+		
+		// Construct html output
+		$html = '';
+		if (!empty($activeTags) || isUser()) {
+			$html.= '<span class="prefix">' . __('tags') . ': </span>';
+			$html.= '<div id="selectedTags" style="display:inline-block;">';
+			if (!empty($selectedTags)) foreach ($selectedTags as $id => $tag) {
```
matthijskooijman: Again, initializing `$selectedTags` to the empty array first removes the need for this if (probably the same for deselected tags).

system/core/tags.php
matthijskooijman: I would write the onclick part as:

`'onclick="deselectTag(' . htmlspecialchars(json_encode($id)).');"'`

This makes sure that any special characters will not mess up the HTML, or the javascript string literal. Note that his uses json_encode to encode a javascript string, which also generates to surrounding quotes.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+				$html.= '<span title="'.$tag['description'].'">'.$tag['label'].'</span>';
+				if (isUser()) $html.= ' <span style="color:#8f0000; cursor:pointer;" title="'.__('remove-tag').'" onclick="deselectTag(\''.$id.'\');">⨯</span>';
+				$html.= '</div>';
+			}
+			$html.= '</div>';
+			if ((isUser() && !empty($deselectedTags)) || isAdmin()) {
+				$html.= '<select id="remainingTags" style="padding:0px;" onchange="selectTag(this.value);">';
+				$html.= '<option value="_add_" selected="selected" disabled="disabled">'.__('add').'...</option>';
+				if (!empty($deselectedTags)) foreach ($deselectedTags as $id => $tag) {
+					$html.= '<option id="tagRem_'.$id.'" value="'.$id.'">'.$tag['label'].'</option>';
```
matthijskooijman: Again, `htmlspecialchars()`.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+		$language = hypha_pageGetLanguage($pageListNode, $lang);
+		$url = $language->getAttribute('id').'/'.$language->getAttribute('name');
+		
+		ob_start();
+?>
+<script>
+	function selectTag(id) {
+		if (!id) addTag();
+		else {
+			hypha('<?=$url?>', 'tagSelect', id, function(response) {document.getElementById('tagList').innerHTML = response;});
```
matthijskooijman: Again, I would use `json_encode` and `htmlspecialchars` here and in similar places below.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+			document.getElementById('tagList').innerHTML = 'loading...';
+		}
+	}
+	function deselectTag(id) {
+		hypha('<?=$url?>', 'tagDeselect', id, function(response) {document.getElementById('tagList').innerHTML = response;});
+		document.getElementById('tagList').innerHTML = 'loading...';
+	}
+	function addTag() {
+		var name = prompt('<?=__('enter-label')?>', '');
```
matthijskooijman: I would use json_encode (and for good measure, also htmlspecialchars though we often leave that out for translations) here and in similar places below.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+	registerCommandCallback('tagNew', 'tagNewCallback');
+	
+	function tagSelectCallback($id) {
+		global $hyphaXml;
+		global $O_O;
+		
+		$lang = $O_O->getContentLanguage();
+		
+		$hyphaXml->lockAndReload();
+		$pageListNode = hypha_getPage($lang, $O_O->getRequest()->getPageName());
```
matthijskooijman: I'm not so sure I like this, since it partially redoes request routing (url to page). The existince of `getPageName()` makes this a lot better than it would be otherwise, though (and I'm not sure if there's a better and also easy way for this...).

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+		$newtag = $hyphaXml->createElement('tag');
+		$newtag->setAttribute('id', $id);
+		$pageListNode->appendChild($newtag);
+		$hyphaXml->saveAndUnlock();
+		
+		echo tagList($pageListNode, $lang);
+		exit;
```
matthijskooijman: I don't like an exit here (since it prevents any cleanup or other processing further up the call stack. Returning some AjaxResponse value might be better, but we don't support that (yet). Might be something to keep in mind for later.

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+		}
+		
+		// Create new tag, add to tagList and page in hypha.xml
+		if (!$tagExists) {
+			$hyphaXml->lockAndReload();
+			$pageListNode = hypha_getPage($lang, $O_O->getRequest()->getPageName());
+			
+			$tagList = $hyphaXml->getElementsByTagName('tagList')->Item(0);
+			$newTag = $hyphaXml->createElement('tag');
+			$newTag->setAttribute('id', $newTagId);
```
matthijskooijman: This should use `$newTag->generateId()` (defined in database.php) rather than generating an id manually. This also means to use the `xml:id` attribute rather than `id` (which allows using getElementsById, IIRC).

system/core/tags.php
```diff
@@ -0,0 +1,184 @@
...
+			$newTagLanguage = $hyphaXml->createElement('language', $newTagDescription);
+			$newTagLanguage->setAttribute('id', $lang);
+			$newTagLanguage->setAttribute('label', $newTagName);
+			$newTag->appendChild($newTagLanguage);
+			$tagList->appendChild($newTag);
+			
+			$newTag = $hyphaXml->createElement('tag');
+			$newTag->setAttribute('id', $newTagId);
```
matthijskooijman: This probably *should* still use `id` (or even something else) rather than `xml:id`, since the latter should be unique within the XML document.

system/core/events.php
giplt: Not sure. In both cases an action is triggered on the client (probably by a user clicking something), and the some data gets routed to a specific piece of code on the server. The only difference is whether the outcome is a fresh page render (by means of a reload) or dynamically handling the outcome in the page that already exists on the client (through ajax). So although the behaviour is somewhat different the purpose (following up on client events) is pretty much the same. From that perspective introducing a separate command adds just as much confusion as it resolves. Maybe up for discussion another time, for now I would argue to adopt this approach. Will make it a separate commit though.

system/core/tags.php
giplt: Kinda intended. Using children would add the necessity to check if the childnode is has 'tag' as tagname. This could be done but adds little in terms of performance or robustness. Lacking a getChildrenByTagName function I chose this approach.

matthijskooijman: This could be easily done using `find` and an xpath expression, I think.

system/core/tags.php
matthijskooijman: We just discussed this and the newtext refactoring will probably significantly change the way commands are handled anyway, so we should probably just leave this as it is now and include this in the newtext design instead.

system/core/tags.php
matthijskooijman: I've added a TODO for later.

system/core/tags.php
matthijskooijman: Laurens fixed this with XPath.

system/core/tags.php
matthijskooijman: Laurens implemented both suggestions.

system/core/tags.php
matthijskooijman: Fixed by Laurens.

system/core/tags.php
matthijskooijman: Fixed by Laurens.

matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: I moved all inline style to the CSS in the theme. I also changed a few "id" attributes to "class", since "id" must be unique within a page, and it does not seem unreasonable for multiple taglists to appear in one HTML page.

I also fixed the escaping.

system/core/tags.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Let's discuss this separately.

system/languages/en.php
matthijskooijman: Do we want to use the word "tag" at all? Or are we going to call these "labels" everywhere anyway (like Github does)?

One alternative could be "Enter label for this tag" (which is what technically happens under the hood - a new tag object is created, which is assigned a label attribute).

system/core/events.php
matthijskooijman: Ok, let's leave it like this for now. We might need to revisit this as part of the newtext refactor anyway, since form handling and command routing is changed there anyway.

system/core/events.php
matthijskooijman: I've moved this, and the addition of the callback parameter to `hypha()`, into seperate commits.

system/core/tags.php
matthijskooijman: I ended up not using `htmlspecialchars` here, since HTML entities are not interpreted inside `<script>` tags. This does leave a small window for problems - if a string value contains `</script>` (or even `</` perhaps), then this could end this script tag. This seems unlikely to happen in practice, though it might be a small security risk. For now, let's just leave it like this (this could perhaps be handled more thoroughly by generating HTML through DomDocument everywhere later).

matthijskooijman:
Laurens and me have done a pass over the code, amending the code. I've force-pushed the result. This resolves all of my remarks. As for my remarks about *where* the code should live and how to structure it, I would suggest postponing that discussion and leaving things as-is now (since the current structure with global functions fits with the rest of Hypha fine).

I think there are a few open point:
 - The comment from Diana above about terminology (Label vs tag).
 - I've changed the code to use `generateId()` and `xml:id`, which is the proper way to define unique ids. However, in some other places, the `id` attribute is still used to *reference* ids (e.g. when referencing a tag id from the pagelist). These cannot be converted to `xml:id` (since they are not unique), but mixing `id` and `xml:id` like this might also be confusing. Perhaps they should be changed to `idref` or just `ref` or something like that? Or more specific `tagid` or `tagref` (though `<tag tagid="..."/>` might also look weird). This might need some face2face discussion.
 - The CSS/layout might need a review, I've not looked at that at all.
 - Any existing customized themes need to be modified to include the `tagList` `div`. Is it time already for some auto-upgrade code for this kind of thing? If so, should it also add CSS automatically?

laurensmartina:
Removed unused `$newTagId` and forced pushed the change.

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
...
+		$selectedTagIds = [];
+
+		/** @var HyphaDomElement $tagEl */
+		// Build list of active tags for given page
+		foreach ($pageListNode->findXPath('tag') as $tagEl) $selectedTagIds[] = $tagEl->getAttribute('id');
+
+		$selectedTags = [];
+		$deselectedTags = [];
+
+		$lang = htmlspecialchars($lang);
```
matthijskooijman: This seems like the wrong place to encode `$lang`. Ususally, you want to encode things as late as possible, just before inserting into (in this case) HTML. Here, `$lang` is encoded to early, and you can see it is then used in an xpath expression (where it should *not* be HTML-encoded, just Xpath-encoded). In fact, it seems that `$lang` is *only* used as part of xpath, never inserted in HTML, so I think this line can simply be dropped?

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
...
+				$html.= '<span title="'.htmlspecialchars($tag['description']).'">'.htmlspecialchars($tag['label']).'</span>';
+				if (isUser()) $html.= ' <span class="delete-tag" title="'.__('remove-tag').'" onclick="deselectTag('.htmlspecialchars(json_encode($id)).');">⨯</span>';
+				$html.= '</div>';
+			}
+			$html.= '</div>';
+			if ((isUser() && !empty($deselectedTags)) || isAdmin()) {
+				$html.= '<select class="remainingTags" onchange="selectTag(this.value);">';
+				$html.= '<option value="_add_" selected="selected" disabled="disabled">'.__('add').'...</option>';
+				foreach ($deselectedTags as $id => $tag) {
+					$html.= '<option id="tagRem_'.htmlspecialchars($id).'" value="'.htmlspecialchars($id).'">'.htmlspecialchars($tag['label']).'</option>';
```
matthijskooijman: What does "Rem" mean in the id here? I would think "Remove", but that's not what this select is for? Remaining then?

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
...
+		$language = hypha_pageGetLanguage($pageListNode, $lang);
+		$url = $language->getAttribute('id').'/'.$language->getAttribute('name');
+
+		ob_start();
+?>
+<script>
+	function selectTag(id) {
+		if (!id) addTag();
+		else {
+			hypha(<?=json_encode($url)?>, 'tagSelect', id, document.getElementById('<?=$formId?>'), function(response) {document.getElementById('tagList').innerHTML = response;});
```
matthijskooijman: I would use `json_encode` for `$formId` here (and below).

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
...
+	function deselectTag(id) {
+		hypha(<?=json_encode($url)?>, 'tagDeselect', id, document.getElementById('<?=$formId?>'), function(response) {document.getElementById('tagList').innerHTML = response;});
+		document.getElementById('tagList').innerHTML = 'loading...';
+	}
+	function addTag() {
+		var name = prompt(<?=json_encode(__('enter-label'))?>, '');
+		if (name) {
+			var description = prompt(<?=json_encode(__('enter-description'))?>, '');
+			if (description) {
+				var json = '{ "name" : "' + name + '", "description" : "' + description + '"}';
```
matthijskooijman: This breaks encoding when name or descriptoin contains a double-quote. I would instead build a JS object and use [JSON.Stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) to encode that to JSON (alternatively, pass the object directly to the `hypha()` function and let that automatically JSON-encode any non-strings or objects it receives).

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
...
+		// Check if tag exists
+		$path = 'hypha/tagList/tag/language[@id=' . xpath_encode($lang) . ' and @label=' . xpath_encode($json['name']) . ']';
+		$tagExists = $hyphaXml->findXPath($path)->count() > 0;
+
+		// Create new tag, add to tagList and page in hypha.xml
+		if (!$tagExists) {
+			$newTagName = $json['name'];
+			$newTagDescription = $json['description'];
+
+			$hyphaXml->lockAndReload();
```
matthijskooijman: This lock should probably happen before checking whether the tag exists (probably even before checking whether the tag list exists), since otherwise there is a race condition where the tag list or tag might be added twice if two requests happen at the same time. This does mean the file is locked for a bit longer, but that should be fine.

system/core/tags.php
```diff
@@ -0,0 +1,192 @@
+<?php
+	// TODO:
+	// - add styling for default theme
```
matthijskooijman: I believe this TODO is resolved with Tammo's commits? The rest below will probably stay until after this first version is merged.

matthijskooijman:
@tammoterhark, there are still quite some open comments that would need to be resolved (or we should decide to not fix them or fix them later, but then we should actually decide that).

system/languages/en.php
matthijskooijman: Conclusion: Let's use "Tag" everywhere, also in the Dutch translation and string ids.

matthijskooijman:
> When adding a tag with an obvious title (example: "Amersfoort" ) the tagging system still needs a description. If you do not supply a description, the tag is not created, no user feedback given.

Conclusion: Make description required, if it is left empty, just ask it again (maybe add a message saying it is required).

tammoterhark:
Datatype shown as datatype.name.tag_list which is possibly unclear for user.

Would prefer: tagpage (since taglist would be a list of tags)

Firefox_Screenshot_2020-02-24T12-07-31 342Z

matthijskooijman:
I've done some more minor fixes, as well as split some commits. @tammoterhark, you had commented the CSS rule for `.prefix` (probably to prevent it from interfering with the taglist), but that changes the languagelist. Instead, I made the `.prefix` rule more specific: b88eb77e7fa41cc195d63e80ea1ed930613bcc11. I also split off some CSS / HTML changes not necessarily to the taglist: 8d8c758b6a19f5398f2912b1009516bc5fd403be and 60d3116be20a43e3c5d03e8a26f1d7fc6363a183. In the future, it would be easier if such unrelated changes would be made in a separate commit directly.

I've also changed terminology to call the tags "tags" rather than label. Note that the terminology "label" and "tag" are both used: A tag *has* a label (and a description). In practice, the label and the tag are usually considered the same thing.

matthijskooijman:
@tammoterhark, your last remark is out of scope for this PR, but I expect that @laurensmartina will solve it before creating a PR for the "taglist" (or what it's called) branch.

Issue #230, opened 2019-03-14
Enable wymeditor alignment and image float plugins


matthijskooijman:
We had these enabled on dewar.nl before the upgrade I'm doing now. Alignment allows setting the text alignment with buttons, image float allows floating images to the side by adding e.g. `float_left` class (so we should add that to the default CSS as well). Image float seemed to work already, alignment didn't when I just tried (but all this was with an older Hypha and thus older wymeditor integration anyway).

matthijskooijman:
There might be a small problem with floating images in the editor. See https://github.com/PlanBCode/hypha/pull/232#issuecomment-472978977 for a workaround.

matthijskooijman:
We should probably also enable the resizable plugin, which makes the editor field resizable by the user.

tammoterhark:
https://bitbucket.org/Patabugen/wymeditor-plugins/src/master/ Could possibly help?

PR #232, opened 2019-03-14
Include regular CSS inside the wymeditor iframe

matthijskooijman:
This makes the WYSYWIG editor a bit more accurate in terms of custom fonts, CSS
classes, etc. This currently somewhat hardcodes the path to the CSS file, which
could do with some refactoring later.

There also seems to be a specific rendering error in the version now running at http://www.dewar.nl (which has this PR applied). There, we use the `float_left` class to make images float. However, in the editor, these are rendered like this:

image

What seems to happen is that the `p` and `hx` tags have `position: relative`, apparently needed to get the 'Heading 4', 'Paragraph`, etc hovering texts, which makes them draw their background over the image somehow. Removing the `position: relative` fixes the image, but also breaks these hover texts.

Additionally (and only marginally related), when you hover over an image, it gets highlighted with some inline CSS. When the hover is removed, this leaves a `margin: 0px` inline CSS rather than removing the inline CSS, which subtly messes up the spacing of the image.

### DIFF
--- a/system/core/editor.php
+++ b/system/core/editor.php
@@ -155,6 +155,14 @@
 				// initiate the "site_links" plugin
 				wym.site_links();
 
+				// Set id so CSS rules apply equally inside the editor as well as outside it
+				$(wym._iframe).contents().find("body").attr('id', 'main');
+
+				// Insert the main css file into the iframe, so it mimics the final result
+				// TODO: Unhardcode
+				var hypha_css = <?= json_encode('data/themes/' . Hypha::$data->theme . '/hypha.css'); ?>;
+				$(wym._iframe).contents().find("head").append($("<link/>", { rel: "stylesheet", href: hypha_css, type: "text/css" }));
+
 				// call the update function when the form is submitted
 				let $form = $(wym.element).closest('form');
 				$form.submit(function(e) {


matthijskooijman:
Turns out this used to work before the update since then we used the wymeditor default theme rather than the pretty theme (which didn't need this `position: relative`). I did some digging and found this is caused because the stacking order is modified by `position: relative`. I found a workaround that I included on www.dewar.nl that fixes this:

```
/* Workaround for a problem with position: relative on p and hx tags in the
 * wymeditor pretty css, that causes these elements (and their white
 * backgrounds) to be drawn over floats rather than under. Making the floats
 * positioned and giving them a z-index puts them later in the stacking order
 * as a workaround. See https://stackoverflow.com/a/43222447/740048
 */
.wym_iframe .float_left { position: relative; z-index: 1; }
.wym_iframe .float_right { position: relative; z-index: 1; }
```

With that covered, I think this could be merged (and if we fix #230 by adding float support in the editor, we might need to include this workaround).

matthijskooijman:
@laurensmartina, could you have a look at this PR?

matthijskooijman:
I tried this on test.destadsbron.nl, and I think this might not really improve things there. On normal pages, this is ok, but the homepage gets quite messed up in the editor.

In particular, the white background that is forced onto paragraphs and other elements does not work too well combined with other background colors:

image

When combined with white text, this is even worse (the text shows up when hovering when it becomes black):

image

Also, the "document division", "paragraph", etc. labels on elements seem to mess up the layout significantly:

image

I think that we could solve this by switching to a simpler wymeditor theme. However, this would mean that for simple pages, the structure offered by the current wymeditor theme is lost, which might be a downside. One could offer a choice between themes in the editor, or maybe in the Hypha site settings (but then it applies to *all* pages, of course)? Reasonable choices would be:
 - Current theme but no site CSS for the most basic view
 - Current theme and site CSS for simple pages that want to use e.g. fonts and simple styling from the site CSS and are not so complex that the visible structure messes things up.
 - Minimal theme and site CSS for the most true WYSYWIG experience.

These wymeditor themes live in the `system/wymeditor/iframe` directory, each directory has a CSS file to be loaded inside the iframe. We're currently using the `pretty` theme, but it seems that both alternatives (`default` and `legacy`) have the same problem (showing structure with overheads and white backgrounds), so for this we would need to create our own theme/iframe structure.

giplt:
I believe this was meant to ease user experience. However it risks to make a opaque mix of the WYSIWYG vs WYSIWYM approaches. Solution should IMHO be in the direction of clarifying these approaches rather than mixing them, by:
- having an explicit explanation embedded in the editor, in a tool tip or otherwise;
- translating the current wording for the functional 'blue blocks' into plain language, and preferably into the user selected interface language, for De Stadsbron simply use "kop", "subkop", "paragraaf", "tabel", "afbeelding" in stead of "heading 4 type no clear" et cetera;
- improving on the preview possibility, e.g. with a split screen containing both the WYM editor and the (instantly rendered) result.

PR #235, opened 2019-03-20
Let (de)wikify handle anchors in links

matthijskooijman:
Before, it would treat the anchor part of a link as part of the
pagename and would create empty pages called e.g. `page_name#anchor`.
Since the links would still be dewikified back to their original value,
the anchors would actually work, but once these dummy pages are deleted,
dewikifying of the links breaks.

To simplify parsing the urls in wikify and dewikify, a regex is used
instead of adding another layer of explode calls.

### DIFF
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -348,14 +348,21 @@
 		global $hyphaContentLanguage;
 		global $hyphaPage;
 
-		$uri = $node->getAttribute('href');
-
-		$path = explode('/', $uri, 2);
-		$parts = explode(':', $path[0], 2);
-		if (count($parts) != 2 || $parts[0] != 'hypha')
+		$href = $node->getAttribute('href');
+		// This matches a url of the form hypha:123abc/subpath#anchor
+		// and splits that into the page id (123abc), followed
+		// by an optional subpath and/or anchor. The page id
+		// will be replaced by the real url, the rest will be
+		// kept as-is.
+		//
+		// Note that pipes are used to enclose the regex rather
+		// than slashes, so we can use slashes without escaping
+		// them.
+		if (!preg_match('|^hypha:([^/#]+)(.*)$|', $href, $matches))
 			return;
+		list($all, $id, $extra) = $matches;
 
-		$page = hypha_getPageById($parts[1]);
+		$page = hypha_getPageById($id);
 		if (!$page)
 			return;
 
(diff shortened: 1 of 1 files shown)


system/core/pages.php
```diff
@@ -347,14 +347,12 @@ function dewikify_link($node) {
 		global $hyphaContentLanguage;
 		global $hyphaPage;
 
-		$uri = $node->getAttribute('href');
-
-		$path = explode('/', $uri, 2);
-		$parts = explode(':', $path[0], 2);
-		if (count($parts) != 2 || $parts[0] != 'hypha')
+		$href = $node->getAttribute('href');
+		if (!preg_match('|^hypha:([^/#]+)(/[^#]*)?(#.*)?$|', $href, $matches))
```
laurensmartina: Regular expressions are, for a lot of people, hard to read. A comment explaining what it does would be helpful.

tammoterhark: I do agree. For non-programmers like me, regex are hell.

matthijskooijman:
I've added some comments, simplified the regex (while explaining it I realized it was overcomplicated) and rebased.

@laurensmartina @tammoterhark, does this comment make things more clear for you>

system/core/pages.php
```diff
@@ -438,12 +444,19 @@ function wikify($elem) {
...
+		//
+		// Note that pipes are used to enclose the regex rather
+		// than slashes, so we can use slashes without escaping
+		// them.
+		if (!preg_match('|^([^/#]+)/([^/#]+)(.*)$|', $href, $matches))
 			return;
-
-		$language = $path[0];
-		$pagename = $path[1];
+		list($all, $language, $pagename, $extra) = $matches;
```
laurensmartina: Seems like duplicate behaviour to `HyphaRequest`. The method in `HyphaRequest` is way more extensive and seems to cover more cases.

matthijskooijman:
Good point, I'll see how we can adapt this to involve HyphaRequest (probably requires some changes to HyphaRequest as well).

matthijskooijman:
@laurensmartina, it turns out using `HyphaRequest` for parsing these urls is a bit more involved and requires changes to `HyphaRequest`. How about we merge this as-is with a regex (since it does fix something right now) with a TODO and then fix this as part of #274?

If you agree, I'll do one more pass over the code, add the TODO and merge it :-)

laurensmartina:
> I've added some comments, simplified the regex (while explaining it I realized it was overcomplicated) and rebased.
> 
> @laurensmartina @tammoterhark, does this comment make things more clear for you?

Yes, it makes it clear.

laurensmartina:
> @laurensmartina, it turns out using `HyphaRequest` for parsing these urls is a bit more involved and requires changes to `HyphaRequest`. How about we merge this as-is with a regex (since it does fix something right now) with a TODO and then fix this as part of #274?
> 
> If you agree, I'll do one more pass over the code, add the TODO and merge it :-)

agreed

PR #243, opened 2019-04-20
Adds a help icon with rules for civil discussion in comments

dianawi:


system/datatypes/peer_reviewed_article.php
```diff
@@ -1158,14 +1158,25 @@ private function createDiscussionCommentForm(DocPage $discussion, array $data =
...
 		$commentFieldName = self::FIELD_NAME_DISCUSSION_COMMENT . '/' . ($new ? 'new_' . $type : $discussion->getId());
+		if (!isUser()) {
+		$html = <<<EOF
+			<div class="new-comment $type">
+			<div>
+				<strong><label for="$commentFieldName"> $comment </label></strong>$infoCommentRules<br/><textarea name="$commentFieldName" id="$commentFieldName" cols="36" rows="4"></textarea>
+			</div>
+EOF;
+		}
+		if (isUser()) {
```
matthijskooijman: I think this if block should be smaller, so there is less duplication of (HTML) code. Probably setting `$infoCommentRule` to the empty string for logged in users is sufficient here.

matthijskooijman:
I made the proposed change. I also moved the info icon into the label tag, so it is shown on the same line as the 'comment on this article` label rather than on a line by itself. I added one more commit to remove a `<br/>` element and I rebased on latest master. Ready to merge now.

Issue #240, opened 2019-04-10
Request/discussion: automatic FaceBook/Twitter posts from pages


tammoterhark:
From the board of De Stadsbron: a wish to automatically post to FaceBook and Twitter.

DISCUSSION?

matthijskooijman:
@tammoterhark is this about automatically posting an article to social media (e.g. when it is published)? Or is this about adding "share" buttons to an article so readers can more easily share articles?

tammoterhark:
This one was about posting the article to our (De Stadsbron's) social media. So only visible when logged in.

This needs discussion about:
- who is allowed to do this
- when can this be done
- how many times can this be done? Is there a need for a counter?
- will we show the text of the article that is placed on social media here?
- How to share images
- will we use Facebook/Twitter specific tools for this?
- ...

PR #244, opened 2019-04-25
Added social media links to peer reviewed article

giplt:
Possible fix for #240 ?
Needs CSS implementation. @tammoterhark: images are now hardcoded to be looked for in the data/images/ folder. Probably you want this to be arranged for in a theme folder.


giplt:
image

giplt:
This feature should probably be implemented as an abstract function (something like 'addMediaLinks') that would be available to all page types. For now I propose to send that to the technical debt department and start testing this feature in peer reviewed article.

tammoterhark:
Please make this also switchable from the admin settings: for The Stadsbron this is now needed, but not for other Hypha instances.

tammoterhark:
No, not a fix for #240, since that is about posting to the own social media by editors of De Stadsbron.

The solution that you implemented is for readers of the Stadsbron.

giplt:
What do you mean by switchable from admin?

tammoterhark:
I mean that in the admin interface (not now, but needed when other parties will want to use this page type) this feature needs to have the possibility to be switched off. Maybe later: switch off/on per page type?

system/datatypes/peer_reviewed_article.php
```diff
@@ -579,6 +579,12 @@ public function indexAction() {
 			$this->findBySelector('#pagename')->before('<div class="number_of_comments">'.$commentCountText.'</div>');
 		}
 
+		$linkToPage = $this->constructFullPath($this->pagename);
+		$shareEmail = '<a href="mailto:?subject='.__('share-email-subject', array("domain" => hypha_getTitle(), "title" => $this->getTitle())).'&body='.__('share-email-body', array("domain" => hypha_getTitle(), "title" => $this->getTitle(), "url" => $linkToPage)).'"><img src="images/shareEmail.png"/></a>';
```
matthijskooijman: I think the subject and body should be urlencoded here.

system/datatypes/peer_reviewed_article.php
```diff
@@ -579,6 +579,12 @@ public function indexAction() {
...
+		$shareTwitter = '<a href="https://twitter.com/intent/tweet?text='.urlencode(__('share-twitter', array("domain" => hypha_getTitle(), "title" => $this->getTitle()))).'&url='.$linkToPage.'" target="_blank"><img src="images/shareTwitter.png"/></a>';
```
matthijskooijman: I think the url= value should be urlencoded.

system/datatypes/peer_reviewed_article.php
```diff
@@ -579,6 +579,12 @@ public function indexAction() {
...
+		$shareFacebook = '<a href="https://www.facebook.com/sharer/sharer.php?u='.$linkToPage.'" target="_blank"><img src="images/shareFacebook.png"/></a>';
```
matthijskooijman: I think the the u= value should be urlencoded here.

matthijskooijman:
> Please make this also switchable from the admin settings:

I think this is a good idea, but I'm not entirely sure how fine-grained this should be. Should you then offer options to enable facebook/twitter/email separately? Should we have options for other datatypes too? Maybe just add this unconditionally for now, and then once we decide how to integrate this in other datatypes too, then decide how we make this optional as well.

matthijskooijman:
Also, this probably interactions with #258, so I think we should merge that and afterwards I can have a look at updating and finishing up this PR.

matthijskooijman:
I've updated this PR to:
 - Fix the encoding issues
 - Move the images into the theme
 - Made the share texts a bit more consistent
 - Removed the "share" text before the icons. Just the icons seems fine, and if we want to add something, we can do that in CSS as well.

image


There's two more things to be done:
 - Add any CSS needed to style this (both in the default theme and for De Stadsbron). Any changes required to the HTML?
 - Confirm that the sharing texts for e-mail and twitter are now ok, or need to be changed.

For the last point:
image

image

image

image

I've put the code of this PR on test.destadsbron.nl, so it can be tested and styled there.

@tammoterhark, can you pick up the above points?

tammoterhark:
I Have added the icons via CSS, so they can be changed by the designer more easily. 

But I also found that the Facebook link gave an error message: Parameter 'href' should represent a valid URL

matthijskooijman:
> But I also found that the Facebook link gave an error message: Parameter 'href' should represent a valid URL

I believe that is because you are testing using a localhost url, which Facebook does not consider valid (it would not make sense to share in any case).

matthijskooijman:
> But I also found that the Facebook link gave an error message: Parameter 'href' should represent a valid URL

I tested on test.destadsbron.nl, works as expected there.

CSS looks good. I've rebased on top of master and will merge in a minute.

Issue #246, opened 2019-05-13
Button for adding comment does not work


tammoterhark:
https://www.destadsbron.nl/nl/Addy_Schuurman_en_Eric_van_der_Velden'

![De Stadsbron (12)](https://user-images.githubusercontent.com/6483317/57621014-ac68d980-758a-11e9-8bf1-535eaa334660.png)

matthijskooijman:
@tammoterhark, have you reproduced this, or are you just forwarding a request from a user? Does it also break on test?

When you click the button, are there any errors in the javascript console?

tammoterhark:
This happens when adding a comment on a comment. Strangely enough the second time I try this, I get a message that two fields are empty, although they were not at the time op submitting this comment-on-comment.

![De Stadsbron (13)](https://user-images.githubusercontent.com/6483317/57621441-b808d000-758b-11e9-8fd8-bc003da88fbf.png)

tammoterhark:
![console](https://user-images.githubusercontent.com/6483317/57621618-37969f00-758c-11e9-81bb-eefaa7aeab3d.png)

tammoterhark:
Error that refers to onloadwff.js is due to one of my plugins.

matthijskooijman:
> This happens when adding a comment on a comment. Strangely enough the second time I try this, I get a message that two fields are empty, although they were not at the time op submitting this comment-on-comment.

This is #200 and, I think, unrelated.

Your javascript errors show that jquery is being blocked, which would explain the problem. I suspect this started happening when HTTPS was enabled (not sure how long ago that was, also not enabled test). It also means that it probably only happens when using https (http is also still available) and it might not happen in all browsers (some browsers only warn and do not block mixed content).

Regardless, I'll have a look at fixing this by simply using https for jquery as well (or rather, use a  // url to use the same protocol as the main page).

matthijskooijman:
Fixed by #248 / ee7fc66580ad7f0009137c4e1b14112edad19560

PR #248, opened 2019-05-13
Use // for external content rather than http://

matthijskooijman:
This fixes mixed content errors when running on HTTPS. In particular,
this prevents jquery from being blocked and breaking significant
content.

Note that this also changes the default footer content as well. Existing
instances of Hypha will continue to use their original footer, which
might still contain an http image link that needs to be changed
manually.


matthijskooijman:
W00ps, forgot to add that this fixes #246.

PR #250, opened 2019-06-11
Remove erronousous appending of port

RoukePouw:


system/core/HyphaRequest.php
```diff
@@ -149,11 +145,7 @@ public function getScheme() {
 
 		public function getHost() {
 			return $_SERVER['HTTP_HOST'];
-		}
-
-		public function getPort() {
-			return $_SERVER['SERVER_PORT'];
-		}
+		}
```
matthijskooijman: This line adds trailing whitespace, it seems (and would not otherwise need to be changed).

system/core/HyphaRequest.php
```diff
@@ -149,11 +145,7 @@ public function getScheme() {
 
 		public function getHost() {
 			return $_SERVER['HTTP_HOST'];
-		}
-
-		public function getPort() {
```
matthijskooijman: I would not remove getPort entirely, other parts of the code might still want to query the port number.

matthijskooijman:
Looks good now, merging!

Issue #92, opened 2018-10-25
Review building self-contained hypha.php


matthijskooijman:
hypha.php allows building a download that contains all relevant files for a new install. I suspect that #85 breaks that feature, probably because files are missing from the generated download. This should be tested and fixed if needed. We might also want to review the way this works right now, since having a whitelist of files to pack might not be ideal. Perhaps git could be involved to get a filelist instead?

PR #252, opened 2019-06-18
Use language.php instead of dummy (only english) language

### LINKED ISSUE #94: Installing from git shows only the English language
When running the hypha.php installer, a list of languages is presented to the user to select the default language. This language list is based on the `languages.php` file contained *inside* `hypha.php`. When running from a downloaded bundle, this file is complete. However, when running from git, the file is already on the filesystem, and there is only a [dummy version](https://github.com/PlanBCode/hypha/blob/124a05cb7ab1756c128c52edb5a11803588b223b/hypha.php#L501) inside hypha.php, to at least not break things.

Ideally, this would simply read the file from the filesystem when it is not present inside `hypha.php`, rather than needing a dummy file, so the complete list of languages is always available. When both versions are available, I think the one from `hypha.php` should take precedence, to allow overwriting an existing install.

RoukePouw:
Fixes #94

Place languages in separate json data file


hypha.php
```diff
@@ -85,7 +85,7 @@ function index($dir = '.') {
 		if ($handle = opendir($dir)) {
 			while (false !== ($file = readdir($handle))) {
 				if ($file != "." && $file != "..") {
-					if (isAllowedFile($dir.'/'.$file)) $index[ltrim($dir.'/'.$file, "./")] = filemtime($dir.'/'.$file);
+					if (isAllowedFile($dir.'/'.$file)) $index[preg_replace("/^\.\//i","",$dir.'/'.$file)] = filemtime($dir.'/'.$file);
```
matthijskooijman: I think this change is unrelated to the rest of the commit (IIRC it fixes installation of the .htaccess file?). Would be better to have this in a separate commit (separate PR is not needed).

hypha.php
```diff
@@ -293,10 +293,8 @@ function install() {
 			if (!is_writable('.')) $errorMessage.= 'php has no write access to the hypa installation directory on the server<br/>';
 
 			// extract list of languages
-			$iso639 = data('system/core/language.php');
-			$p1 = strpos($iso639, "json_decode('");
-			$p2 = strpos($iso639, "', true);", $p1);
-			$iso639 = json_decode(preg_replace('/\\\\\'/', '\'', substr($iso639, $p1+13, $p2-$p1-13)));
+			$iso639 = data('system/languages/languages.json');
+			$iso639 = json_decode(preg_replace('/\\\\\'/', '\'', $iso639));
```
matthijskooijman: Why do we still need this replacement? Isn't the content of `languages.json` valid json as-is?

hypha.php
```diff
@@ -420,7 +418,9 @@ function data($name) {
 		switch ($name) {
 			//START_OF_DATA
 			case 'index': return '';
-			case 'system/core/language.php': return "json_decode('{\"en\": \"English\"}', true);";
+			case 'system/core/language.php':
+        require_once('system/core/language.php');
+        return $isoLangList;
```
matthijskooijman: Shouldn't this have a case for `languages.json` rather than `language.php` now? And that could just return the `file_get_contents()` for the existing `languages.json`?

system/core/language.php
```diff
@@ -33,4 +33,9 @@ function languageOptionList($select, $omit) {
...
+  
+  if(file_exists('system/languages/languages.json')){
+    $isoLangList = json_decode(file_get_contents('system/languages/languages.json'), true);
+  }else{
```
matthijskooijman: Do we need this check and else? AFAICS we always have a `languages.json` whenever `languages.php` also exists, so we could just remove the if?

hypha.php
matthijskooijman: Ah, it seems these `\'` is present in the `languages.json` file you added. These were probably needed when the json was contained in a PHP `''` string, but this is no longer relevant. I would rather remove these backslashes from the `languages.json` file and remove the replacement above.

hypha.php
matthijskooijman: This commit could probably be set to close #92.

matthijskooijman: This comment is still unresolved.

hypha.php
matthijskooijman: This is resolved.

system/core/language.php
matthijskooijman: This is resolved.

hypha.php
matthijskooijman: This is partly resolved, but I'll add a new comment for what's left.

hypha.php
```diff
@@ -420,7 +418,7 @@ function data($name) {
 		switch ($name) {
 			//START_OF_DATA
 			case 'index': return '';
-			case 'system/core/language.php': return "json_decode('{\"en\": \"English\"}', true);";
+			case 'system/languages/languages.json': return json_decode(file_get_contents('system/languages/languages.json'), true);
```
matthijskooijman: This now returns the decoded json (e.g. a PHP array), whereas the version that is generated when packing up hypha has this section replaced with actual file data, so that changes the return value. IOW, when you pack up hypha, the code no longer works.

To fix that, this should just return the result of `file_get_contents()`, and the `json_decode()` should be moved to the caller of `data()` elsewhere. Might be good to also test this packing up (though I'm not 100% sure it is working in master).

laurensmartina: fixed

hypha.php
laurensmartina: done

Issue #146, opened 2018-11-28
New page dialog includes "settings" pagetype


matthijskooijman:
This is probably a side effect of a697fc2a56b09ebec3cfe48231b608ff63493d8d

I'm not sure what the best fix is. Perhaps settings.php should be moved out of datatypes, since it is not really a datatype, only a (special) page type I believe?

giplt:
I suggest it to be a reserved datatype name and hardcode its exclusion from the new page dialog

PR #254, opened 2019-06-18
Feature: Make default page type configurable

### LINKED ISSUE #151: Feature: Make default page type configurable
The default page type (e.g. the datatype that is preselected when you click "new page") is now (I believe) hardcoded to be webpage (text). However, some instances, use another type more intensively (e.g. article in the stadsbron) and it would be helpful to users to preselect that type. This could be simply a pagetype dropdown in the settings page (similar to the default language).

#137 also suggests that such a default pagetype could be expanded into an orderable list of pagetypes (so you can change the default, but also the order), but that would require more effort for less gain, so it seems better to implement just the default selection first.

RoukePouw:
Fixes #151

### DIFF
--- a/index.php
+++ b/index.php
@@ -39,18 +39,29 @@
 		Group: Stage 2 - Build hypha script
 		Additional scripts and data are included.
 		- *core* Include core scripts
-		- *dataypes* Include modules for available datatypes: textpage, mailinglist, blog et cetera
 		- *languages* Compile a list of available user interface languages
+		- *dataypes* Include modules for available datatypes: textpage, mailinglist, blog et cetera
 	*/
 
 	foreach (scandir("system/core/") as $_file) if (substr($_file, -4) == '.php') require_once("system/core/" . $_file);
+
+	// Build request
+	$hyphaRequest = new HyphaRequest($isoLangList);
+
+	// Build request context
+	$O_O = new RequestContext($hyphaRequest, hypha_getDefaultLanguage());
+
+	foreach (scandir("system/languages/") as $_file) if (substr($_file, -4) == '.php') $uiLangList[] = basename($_file, '.php');
 
 	$pageTypeIndex = count(get_declared_classes()) -1;
 	foreach (scandir("system/datatypes/") as $_file) if (substr($_file, -4) == '.php') include_once("system/datatypes/" . $_file);
 	$hyphaPageTypes = [];
 	foreach (array_slice(get_declared_classes(), $pageTypeIndex) as $class) if (in_array('Page', class_parents($class)) && (new \ReflectionClass($class))->isInstantiable()) $hyphaPageTypes[] = $class;
 
-	foreach (scandir("system/languages/") as $_file) if (substr($_file, -4) == '.php') $uiLangList[] = basename($_file, '.php');
+	$dataTypeMtx = [];
+	foreach ($hyphaPageTypes as $className) {
+		$dataTypeMtx[$className] = call_user_func($className . '::getDatatypeName');
+	}
 
 	/*
 		Group: Stage 3 - Get query and handle direct file requests
@@ -58,8 +69,6 @@
 		Some requests don't need the whole script to run. Certain file requests (css, javascript) can be handled directly here.
 	*/
 
-	// Build request
-	$hyphaRequest = new HyphaRequest($isoLangList);
 	$hyphaQuery = $hyphaRequest->getRelativeUrlPath();
 	$hyphaUrl = $hyphaRequest->getRootUrl();
 
@@ -95,9 +104,6 @@
 		Group: Stage 5 - Initialize user and determine language to use
 		Check is a user is logged in and load user data. Add login/logout functionality according to session login status.
 	*/
-
-	// Build request context
-	$O_O = new RequestContext($hyphaRequest, hypha_getDefaultLanguage());
 
 	/**
 	 * Set user as global variable.
@@ -141,10 +147,6 @@
 		$_cmds[] = '<a href="javascript:login();">'.__('login').'</a>';
 	}
 	else {
-		$dataTypeMtx = [];
-		foreach ($hyphaPageTypes as $className) {
-			$dataTypeMtx[$className] = call_user_func($className . '::getDatatypeName');
-		}
 		addNewPageRoutine($hyphaHtml, $hyphaRequest->getRelativeUrlPathParts(false), $dataTypeMtx);
 		$_cmds[] = makeLink(__('new-page'), 'newPage();');
 		$_cmds[] = makeLink(__('settings'), makeAction('settings', '', ''));
--- a/index.php
+++ b/index.php
@@ -53,15 +53,8 @@
 
 	foreach (scandir("system/languages/") as $_file) if (substr($_file, -4) == '.php') $uiLangList[] = basename($_file, '.php');
 
-	$pageTypeIndex = count(get_declared_classes()) -1;
-	foreach (scandir("system/datatypes/") as $_file) if (substr($_file, -4) == '.php') include_once("system/datatypes/" . $_file);
-	$hyphaPageTypes = [];
-	foreach (array_slice(get_declared_classes(), $pageTypeIndex) as $class) if (in_array('Page', class_parents($class)) && (new \ReflectionClass($class))->isInstantiable()) $hyphaPageTypes[] = $class;
-
-	$dataTypeMtx = [];
-	foreach ($hyphaPageTypes as $className) {
-		$dataTypeMtx[$className] = call_user_func($className . '::getDatatypeName');
-	}
+	// called for its side effects of loading the datatypes
+	hypha_getDataTypes();
 
 	/*
 		Group: Stage 3 - Get query and handle direct file requests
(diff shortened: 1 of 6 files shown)


system/datatypes/settings.php
```diff
@@ -218,7 +218,28 @@ function register($key) {
...
+		 Function: typesOptionList
+		 Returns an html option list with data types
+		 
+		 Parameters:
+		 $select - which option should be preselected
+		 $omit - which data types should be omitted
+		 */
+		function typesOptionList($select, $omit) {
+			$html = '';
+			foreach (scandir('system/datatypes') as $file) {
```
matthijskooijman: I think that this does not match the way the new page dialog generates the list of possible page types, see index.php and the `$hyphaPageTypes` variable. It might be good to move the list of known page types to some more centralized position, rather than generating it inside `index.php` so it can be reused here. Doing so might also make fixing #146 easier perhaps.

system/languages/nl.php
```diff
@@ -50,6 +50,7 @@
 		"default-page" => "standaard pagina",
 		"default-language" => "standaard taal",
 		"digest-interval" => "notificatie-interval",
+		"default-new-page-type" => "standaard type voor nieuwe pagina",
```
matthijskooijman: `standaardtype` is one word, I think :-)

system/languages/en.php
```diff
@@ -50,6 +50,7 @@
 		"default-page" => "default page",
 		"default-language" => "default language",
 		"digest-interval" => "digest interval",
+		"default-new-page-type" => "new page type",
```
matthijskooijman: Did you intentionally leave out "default" here? To mimic the Dutch version, perhaps "default type for new pages" might be better?

system/languages/nl.php
matthijskooijman: Also, I would say "pagina's" rather than "pagina".

system/datatypes/settings.php
RoukePouw: "some more centralized position," any suggestions?

system/languages/nl.php
RoukePouw: Fixed

system/languages/en.php
RoukePouw: Not intentionally, fixed

system/datatypes/settings.php
```diff
@@ -218,7 +218,28 @@ function register($key) {
...
+				if ('.' === $file) continue;
+				if ('..' === $file) continue;
+				$name = substr($file,0,-4);
+				if($name!=$omit){
+					$html.= '<option value='.$name.($name==$select ? ' selected' : '').'>'.$name.'</option>';
+				}
+			}
+			return addslashes($html);
```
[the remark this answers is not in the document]
matthijskooijman: Hm, it seems that `addSlashes()` has no real purpose there either and would be better replaced by `htmlspecialchars()` in appropriate places. That's for a different PR, but let's make this new code correct right away :-)

I would suggest:

    $html.= '<option value='.htmlspecialchars($name).($name==$select ? ' selected' : '').'>'.htmlspecialchars($name).'</option>';

and then remove the `addSlashes()` call.

system/datatypes/settings.php
matthijskooijman: There is a global `$hyphaPageTypes` that is documented in `base.php`, so I guess that would be a good place. I can imagine a `hypha_getPageTypes()` function in that same file that returns that variable? Though it seems you might not just need the (class)names, but also the "titles" (display names) of the page types (which you now assume are equal, [but apparently are not](https://github.com/PlanBCode/hypha/blob/master/index.php#L146)).

Perhaps `$hyphaPageTypes` could be extended to be an array of `classname => displayname` (essentially what [`$dataTypeMtx` in `index.php`] already is), or the new `hypha_getPageTypes()` could be extended to load the display names on-demand?

Adding such a function is probably best done in a separate commit, since it is essentially an independent change.

Maybe it would also be good to move the code that fills this variable from `index.php` into `base.php` as well (e.g. into a `hypha_findPageTypes()` or something?). That might be out of scope, but if we include the display name stuff now, that might involve refactoring index.php anyway, so perhaps better to do this properly right away?

system/core/base.php
```diff
@@ -801,3 +827,28 @@ function hypha_substitute($string, array $vars) {
...
+				if (in_array(Page::class, class_parents($class)) && (new ReflectionClass($class))->isInstantiable()) {
+					$hyphaPageTypes[$class] = call_user_func($class . '::getDatatypeName');
+				}
+			}
+		}
+
+		$dataTypes = $hyphaPageTypes;
+		$systemDataTypes = [settingspage::class];
+
+		if ($excludeSystemDataTypes) foreach ($systemDataTypes as $systemDataType) unset($dataTypes[$systemDataType]);
```
matthijskooijman: AFAICS, the excludeSystemDataTypes is always passed as true. Do we really need to default to false, or even have this parameter in the first place?

index.php
```diff
@@ -45,12 +45,11 @@
 
 	foreach (scandir("system/core/") as $_file) if (substr($_file, -4) == '.php') require_once("system/core/" . $_file);
 
-	$pageTypeIndex = count(get_declared_classes()) -1;
-	foreach (scandir("system/datatypes/") as $_file) if (substr($_file, -4) == '.php') include_once("system/datatypes/" . $_file);
-	$hyphaPageTypes = [];
-	foreach (array_slice(get_declared_classes(), $pageTypeIndex) as $class) if (in_array('Page', class_parents($class)) && (new \ReflectionClass($class))->isInstantiable()) $hyphaPageTypes[] = $class;
+	// Build request context
+	$O_O = new RequestContext(new HyphaRequest($isoLangList), hypha_getDefaultLanguage());
```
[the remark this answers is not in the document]
laurensmartina: done

system/core/base.php
```diff
@@ -801,3 +827,28 @@ function hypha_substitute($string, array $vars) {
 		foreach ($vars as $key => $val) $string = str_replace('[[' . $key . ']]', $val, $string);
 		return $string;
 	}
+
+	/*
+		Function: getHyphaPageTypes
```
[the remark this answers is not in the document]
laurensmartina: done

system/core/base.php
```diff
@@ -801,3 +827,28 @@ function hypha_substitute($string, array $vars) {
...
+		Returns an list with data types
+	*/
+	function getHyphaDataTypes($excludeSystemDataTypes = false) {
```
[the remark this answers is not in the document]
laurensmartina: done

system/core/base.php
laurensmartina: It is used without a parameter in the `index.php`

system/core/base.php
```diff
@@ -801,3 +827,28 @@ function hypha_substitute($string, array $vars) {
...
+		global $hyphaPageTypes;
+		if (empty($hyphaPageTypes)) {
+			$hyphaPageTypes = [];
+			$pageTypeIndex = count(get_declared_classes()) -1;
+			foreach (scandir('system/datatypes/') as $file) if (substr($file, -4) == '.php') include_once('system/datatypes/' . $file);
```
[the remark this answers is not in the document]
laurensmartina: It is called from `index.php`, now in a separate commit.

system/datatypes/settings.php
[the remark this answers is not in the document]
laurensmartina: done

system/datatypes/settings.php
laurensmartina: done

index.php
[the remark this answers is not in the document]
laurensmartina: done

system/core/base.php
```diff
@@ -801,3 +827,34 @@ function hypha_substitute($string, array $vars) {
 		foreach ($vars as $key => $val) $string = str_replace('[[' . $key . ']]', $val, $string);
 		return $string;
 	}
+
+	/*
+		Function: hypha_getDataTypes
+		Returns an list with dataTypes
+	*/
+	function hypha_getDataTypes($excludeSystemDataTypes = false) {
```
matthijskooijman: In a (resolved) previous comment, I said:

>AFAICS, the excludeSystemDataTypes is always passed as true. Do we really need to default to false, or even have this parameter in the first place?

Laurens said:

> It is used without a parameter in the index.php

This is true, but that actually means that the "new page" dialog (still) offers to make a `settingspage` (which is #146). I think we can just drop the exclude parameter so the function will always:
 - As a side effect, load *all* datatypes (if not done before).
 - Return only non-settings datatypes.

matthijskooijman:
I expect this will also slightly conflict with #280 and/or #252, so some care must be taken when merging.

system/core/base.php
```diff
@@ -801,3 +801,26 @@ function hypha_substitute($string, array $vars) {
 		foreach ($vars as $key => $val) $string = str_replace('[[' . $key . ']]', $val, $string);
 		return $string;
 	}
+
+	/*
+		Function: hypha_getDataTypes
+		Returns an list with dataTypes
+	*/
+	function hypha_getDataTypes() {
+		global $hyphaDataTypes;
```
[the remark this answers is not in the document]
laurensmartina: I never knew you could define a static variable outside a class, very useful.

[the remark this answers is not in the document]
laurensmartina: fixed

index.php
```diff
@@ -52,16 +52,7 @@
...
-	$pageTypeIndex = count(get_declared_classes()) -1;
-	foreach (scandir("system/datatypes/") as $_file) if (substr($_file, -4) == '.php') include_once("system/datatypes/" . $_file);
-	$hyphaPageTypes = [];
-	foreach (array_slice(get_declared_classes(), $pageTypeIndex) as $class) if (in_array('Page', class_parents($class)) && (new \ReflectionClass($class))->isInstantiable()) $hyphaPageTypes[] = $class;
-
-	$dataTypeMtx = [];
-	foreach ($hyphaPageTypes as $className) {
-		$dataTypeMtx[$className] = call_user_func($className . '::getDatatypeName');
-	}
+	$hyphaDataTypes = hypha_getDataTypes();
```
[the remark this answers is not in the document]
laurensmartina: fixed

system/core/base.php
laurensmartina: done

matthijskooijman:
Looks good like this. I fixed one more whitespace issue and rebased on top of master, merging next.

PR #260, opened 2019-06-25
Suggestion for handling errors.

laurensmartina:
With the merge of newtext an 'errors' return has been introduced, but this return type is not being handled. This pull request is to start the discussion on how to handle this new return type.

### DIFF
--- a/system/core/events.php
+++ b/system/core/events.php
@@ -328,6 +328,13 @@
 			header('Location: ' . $url);
 			exit;
 		}
+
+		if (is_array($result) && array_key_exists('errors', $result)) {
+			foreach ($result['errors'] as $error) {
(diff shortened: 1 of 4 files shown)


matthijskooijman:
I wondered for a bit whether we really need this (the code can also directly call notify and return), but it might be better to do this centrally.

The basic structure looks good to me. I do wonder if we should use notify here and render the page normally, or whether we should just render a full-screen error message instead of the normal page (this would also be good for 404 errors). I also wonder what to do with errors that are generated in POST request (such as this one). Ideally, you would redirect to a GET request and then show the error, but then you need to preserve the error message (which already works for notify, which would conflict with the thought of not using notify).

Just some initial thoughts :-)

PR #276, opened 2019-11-25
Fix PHP5: Remove return type declarations from php-dom-wrapper

matthijskooijman:
These were added in commit 407c83a (Backport php-dom-wrapper's create,
appendTo and prependTo methods), unintentionally raising the minimum PHP
version to 7.0. By removing these return type declarations, the code
should be compatible with PHP 5 again.


matthijskooijman:
Rebased and merged.

Issue #269, opened 2019-10-31
Add "(TEST)" in newsletter title when sent in testmode


dianawi:
when a newsletter is sent to testers, they can not tell if it is the final newsletter or a draft that they have to check.

dianawi:
suggested text to add in the subject: testnieuwsbrief, graag controleren!/ newsletter test, please check!

dianawi:
Done by Laurens, no pull request yet

matthijskooijman:
According the the PR and the project board, this was done in #277, so closing this issue.

PR #277, opened 2019-11-25
mailinglist: Updated to new standard

laurensmartina:


system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
 			notify('error', __('missing-arguments'));
-
 			return null;
 		}
 
 		$this->xml->lockAndReload();
 
 		/** @var NodeList $addressCollection */
-		$addressCollection = $this->getDoc()->findXPath('//address[@confirm-code="' . $code . '"]');
+		$xpath = '//' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_CONFIRM_CODE . '="' . $code . '" and @' . self::FIELD_NAME_STATUS . ' != "' . self::ADDRESS_STATUS_UNSUBSCRIBED . '"]';
```
matthijskooijman: this should also encode xpath

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
 			return ['redirect', $this->constructFullPath($this->pagename)];
 		}
 
-		$address->setAttribute('status', self::ADDRESS_STATUS_CONFIRMED);
-		$address->setAttribute('unsubscribe-code', $this->constructCode());
+		$address->setAttribute(self::FIELD_NAME_STATUS, self::ADDRESS_STATUS_CONFIRMED);
+		$address->setAttribute(self::FIELD_NAME_UNSUBSCRIBE_CODE, $this->constructCode());
 		/** @var HyphaDomElement $addresses */
-		$addresses = $this->getDoc()->getOrCreate('addresses');
+		$addresses = $this->xml->find(self::FIELD_NAME_ADDRESSES_CONTAINER);
```
matthijskooijman: Should this use find rather than getOrCreate? With find, it will look recursively, which is not what is intended I think? Maybe this should use `get()` instead?

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
-
 			return null;
 		}
 
 		$this->xml->lockAndReload();
+
 		/** @var NodeList $addressCollection */
-		$addressCollection = $this->getDoc()->findXPath('//address[@unsubscribe-code="' . $code . '"]');
+		$xpath = '//' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_UNSUBSCRIBE_CODE . '=' . xpath_encode($code) . ']';
+		$addressCollection = $this->xml->findXPath($xpath);
```
matthijskooijman: Should this find be limited in scope to the addresses element? Maybe alos use './' to only get direct children rather than look recursively? This also applies to other findXPaths in the code.

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
-
 			return null;
 		}
 
-		if ($address->getAttribute('status') != self::ADDRESS_STATUS_UNSUBSCRIBED) {
-			$address->setAttribute('status', self::ADDRESS_STATUS_UNSUBSCRIBED);
-			$address->removeAttribute('email');
+		if ($address->getAttribute(self::FIELD_NAME_STATUS) != self::ADDRESS_STATUS_UNSUBSCRIBED) {
+			$address->setAttribute(self::FIELD_NAME_STATUS, self::ADDRESS_STATUS_UNSUBSCRIBED);
+			$address->setAttribute(self::FIELD_NAME_UNSUBSCRIBE_EMAIL, $address->getAttribute(self::FIELD_NAME_EMAIL));
```
matthijskooijman: What is the idea behind this extra field? If we want to keep the e-mail address, why not just in the regular e-mail field?

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
-			if (null == $mailingId) {
-				$mailingId = $_POST[self::FIELD_NAME_ID];
-			} elseif ($_POST[self::FIELD_NAME_ID] != $mailingId) {
-				notify('error', __('ml-inconsistent-parameters'));
 
-				return null;
+		// display mailing
+		/** @var HyphaDomElement $main */
+		$main = $this->html->find('#main');
+		$main->append('<div><h2>' . $mailing->getAttribute('subject') . '</h2></div>');
```
matthijskooijman: htmlspecialchars

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
+		// add edit button when in draft
+		if (self::MAILING_STATUS_DRAFT == $status) {
+			$path = str_replace('{id}', $mailing->getId(), self::PATH_MAILS_EDIT_ID);
+			$commands->append($this->makeActionButton(__('edit'), $path));
+			$xpath = '//' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_STATUS . '="' . self::ADDRESS_STATUS_CONFIRMED . '"]';
+			$num = count($this->xml->findXPath($xpath));
+
+			$path = str_replace('{id}', $mailing->getId(), self::PATH_MAILS_VIEW_ID);
+			$path = $this->language . '/' . $this->pagename . '/' . $path;
+			$commands->append(makeButton(__('send'), 'if(confirm(\'' . __('ml-sure-to-send', array("count" => $num)) . '\'))' . makeAction($path, self::CMD_SEND, '')));
```
matthijskooijman: This only adds the send button when in draft, is that intentional?

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
+			$commands->append(makeButton(__('ml-test-send'), 'hypha(\''.$path.'\', \''.self::CMD_SEND_TEST.'\', prompt(\'' . __('email') . '\'), $(this).closest(\'form\'));'));
```
matthijskooijman: I would use `json_encode()` for all these javascript strings.

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
-			}
+		// add edit button when in draft
+		if (self::MAILING_STATUS_DRAFT == $status) {
+			$path = str_replace('{id}', $mailing->getId(), self::PATH_MAILS_EDIT_ID);
+			$commands->append($this->makeActionButton(__('edit'), $path));
+			$xpath = '//' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_STATUS . '="' . self::ADDRESS_STATUS_CONFIRMED . '"]';
+			$num = count($this->xml->findXPath($xpath));
+
+			$path = str_replace('{id}', $mailing->getId(), self::PATH_MAILS_VIEW_ID);
+			$path = $this->language . '/' . $this->pagename . '/' . $path;
```
matthijskooijman: I think this can use `constructFullPath()`?

system/datatypes/mailinglist.php
```diff
@@ -660,116 +785,27 @@ private function createEditForm(array $data = []) {
...
-		$subject = __('subject');
-		$subjectFieldName = self::FIELD_NAME_SUBJECT;
-		$message = __('message');
-		$messageFieldName = self::FIELD_NAME_MESSAGE;
 		$html = <<<EOF
-			<input type="hidden" id="$idFieldName" name="$idFieldName" />
+			<input type="hidden" id="[[field-name-id]]" name="[[field-name-id]]" />
 			<div class="section" style="padding:5px; margin-bottom:5px; position:relative;">
-				<label for="$subjectFieldName"> $subject </label> <input type="text" id="$subjectFieldName" name="$subjectFieldName" />
+				<field-name-labelsubjectFieldName]]"> [[subject]] </field-name-label> <input type="text" id="[[subjectFieldName]]" name="[[subjectFieldName]]" />
```
matthijskooijman: This looks broken: `<field-name-labelsubjectFieldName]]">`

Also, `</field-name-label>` is not valid HTML tag? What's going  on here?

Also, `[[subjectFieldName]]` should be `[[field-name-subject]]` I think?

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
 		}
-		if (isUser()) {
-			$receivers = $mailing->getAttribute('receivers');
-			if (null != $receivers) {
-				$main->append('<div>' . __('ml-received-by') . ': ' . $receivers . '</div>');
-			}
+
+		$postedId = $request->getPostValue(self::FIELD_NAME_ID);
+		if ($postedId != $mailingId) {
+			notify('error', __('ml-inconsistent-parameters'));
```
matthijskooijman: Why do we include the mailing id in the POST parameters at all? Only for this consistency check it seems? Why not leave it out of the form entirely? Originally, it seems it was used to distinguish between new mailings and edited mailings, but this is now done using arguments to helper functions, so just putting the id in the URL seems sufficient?

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
 		}
-		$main->append($mailing->getHtml());
 
-		// buttons
-		$commands = $this->findBySelector('#pageCommands');
-		$commands->append($this->makeActionButton(__('back'), ''));
+		// check if status is correct
+		$status = $mailing->getAttribute('status');
+		if (self::MAILING_STATUS_SENT == $status) {
+			notify('success', __('ml-successfully-sent'));
```
matthijskooijman: Huh? When status is SENT, this throes away the edit and says "successfully sent"? What's the idea here?

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
-		/** @var \DOMWrap\Element $elem */
-		$elem = $form->html($html);
 
-		// buttons
-		/** @var NodeList $field */
-		$field = $elem->find('#' . $emailFieldName);
-		$field->parent()->append($this->makeActionButton(__('subscribe'), null, self::FORM_CMD_LIST_SUBSCRIBE));
+		$vars = [
+			'email' => __('email'),
+			'emailFieldName' => self::FIELD_NAME_EMAIL,
```
matthijskooijman: This should be `field-name-email`.

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
+		];
+
+		$html = hypha_substitute($html, $vars);
 
-		return $this->createForm($elem->children(), $data);
+		return new WymHTMLForm($html, $values);
```
matthijskooijman: This can be a normal `HTMLForm`.

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
+			'senderName' => __('mailing-sender-name'),
+			'senderNameFieldName' => self::FIELD_NAME_SENDER_NAME,
+			'senderEmail' => __('mailing-sender-email'),
+			'senderEmailFieldName' => self::FIELD_NAME_SENDER_EMAIL,
+			'privatePage' => __('private-page'),
+			'privateFieldName' => self::FIELD_NAME_PRIVATE,
+			'description' => __('description'),
+			'descriptionFieldName' => self::FIELD_NAME_DESCRIPTION,
+			'emailWelcomeText' => __('email-welcome-text'),
+			'emailWelcomeTextFieldName' => self::FIELD_NAME_EMAIL_WELCOME_TEXT,
```
matthijskooijman: These should also be `field-name-xxx`.

system/datatypes/mailinglist.php
```diff
@@ -781,52 +817,111 @@ private function testSendMailing($mailingId, $email) {
...
+	 * @param HyphaRequest $request
 	 *
 	 * @throws Exception
+	 * @return null
 	 */
-	private function sendMailing($mailingId) {
+	private function mailingSendAction(HyphaRequest $request) {
 		if (!$this->hasSender()) {
-			throw new \Exception(__('ml-no-sender'));
+			notify('error', __('login-to-edit'));
```
matthijskooijman: This mixes up two checks: `!hasSender()` now shows `login-to-edit` (which is wrong), and there is no actual `!isUser()` check (seems it was not there before either).

system/datatypes/mailinglist.php
```diff
@@ -781,52 +817,111 @@ private function testSendMailing($mailingId, $email) {
...
+
+		// get status
+		$status = $mailing->getAttribute('status');
+		if (self::MAILING_STATUS_DRAFT !== $status) {
+			notify('error', __('ml-mail-has-invalid-status'));
+			return null;
+		}
+
+		$senderName = $this->xml->documentElement->getAttribute('sender-name');
+		$senderEmail = $this->xml->documentElement->getAttribute('sender-email');
```
matthijskooijman: Shouldn't this use getSenderData?

system/datatypes/mailinglist.php
```diff
@@ -781,52 +817,111 @@ private function testSendMailing($mailingId, $email) {
...
+		$receivers = $this->xml->findXPath($xpath);
 		$linkToMailing = $this->constructFullPath($this->pagename . '/' . $mailingId);
 		foreach ($receivers as $receiver) {
 			$email = $receiver->getAttribute('email');
 			$code = $receiver->getAttribute('unsubscribe-code');
-			$linkToUnsubscribe = $this->constructFullPath(sprintf('%s/unsubscribe?address=%s&code=%s', $this->pagename, $receiver->getId(), $code));
+			$linkToUnsubscribe = $this->constructFullPath($this->pagename . '/' . str_replace(['{address}', '{code}'], [$receiver->getId(), $code], self::PATH_UNSUBSCRIBE_CODE));
 			$message = '<p><a href="' . $linkToMailing . '">' . __('ml-if-unreadable-use-link') . '</a></p>';
 			$message .= $mailing->getHtml();
 			$message .= '<p><a href="' . $linkToUnsubscribe . '">' . __('ml-unsubscribe') . '</a></p>';
```
matthijskooijman: htmlspecialchars on links

system/datatypes/mailinglist.php
```diff
@@ -781,52 +817,111 @@ private function testSendMailing($mailingId, $email) {
...
+			return null;
+		}
+
+		$email = $request->getPostValue('argument');
+		$code = 'this-is-your-unsubscribe-code';
+		$linkToMailing = $this->constructFullPath($this->pagename . '/' . $mailingId);
+		$linkToUnsubscribe = $this->constructFullPath($this->pagename . '/' . str_replace(['{address}', '{code}'], [$email, $code], self::PATH_UNSUBSCRIBE_CODE));
+		$message = '<p><a href="' . $linkToMailing . '">' . __('ml-if-unreadable-use-link') . '</a></p>';
+		$message .= $mailing->getHtml();
+		$message .= '<p><a href="' . $linkToUnsubscribe . '">' . __('ml-unsubscribe') . '</a></p>';
```
matthijskooijman: htmlspecialchars on links

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
-	private $xml;
-
 	/**
 	 * @param DOMElement $pageListNode
 	 * @param RequestContext $O_O
 	 */
 	public function __construct(DOMElement $pageListNode, RequestContext $O_O) {
 		parent::__construct($pageListNode, $O_O);
-		$this->xml = new Xml('mailinglist', Xml::multiLingualOn, Xml::versionsOff);
+		$this->xml = new Xml(get_called_class(), Xml::multiLingualOn, Xml::versionsOff);
```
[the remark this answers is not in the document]
laurensmartina: updated it

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
-			$mailings = $this->getDoc()->findXPath('//mailing[@status="' . self::MAILING_STATUS_SENT . '"]');
+	/**
+	 * Deletes the mailing list.
+	 *
+	 * @param HyphaRequest $request
+	 * @return array
+	 */
+	public function deleteAction(HyphaRequest $request) {
+		if (!isAdmin()) {
+			return ['errors' => [__('insufficient-rights-to-perform-action')]];
```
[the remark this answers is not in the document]
laurensmartina: true, i'll change this

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
+	 * @return string|null
+	 */
+	public function defaultView(HyphaRequest $request) {
+		// create form
+		$form = $this->createSubscribeForm();
+
+		return $this->indexViewRender($request, $form);
+	}
 
+	private function indexViewRender(HyphaRequest $request, WymHTMLForm $form) {
```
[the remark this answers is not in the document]
laurensmartina: changed it

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
+			return $this->indexViewRender($request, $form);
+		}
+
+		// check if email is already in the list, no need to have it twice.
+		$this->xml->lockAndReload();
+		$email = strtolower($form->dataFor(self::FIELD_NAME_EMAIL));
+		/** @var HyphaDomElement $addresses */
+		$addresses = $this->xml->find(self::FIELD_NAME_ADDRESSES_CONTAINER);
+		/** @var NodeList $addressCollection */
+		$xpath = '//' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_EMAIL . '="' . $email . '"]';
```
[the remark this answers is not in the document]
laurensmartina: changed it

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
+		$addressCollection = $addresses->children()->findXPath($xpath);
```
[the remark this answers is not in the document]
laurensmartina: It's tested and now tested again, so it does work. That said, I'll look into this.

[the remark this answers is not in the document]
laurensmartina: Removing `->children()` also works. :-)

system/datatypes/mailinglist.php
```diff
@@ -1,406 +1,433 @@
...
+			$this->xml->saveAndUnlock();
+		} else {
+			$this->xml->unlock();
+		}
+
+		if ($sendEmail) {
+			// send email so that pending subscribed can confirm email address
+			$confirmPath = str_replace('{code}', $code, self::PATH_CONFIRM_CODE);
+			$confirmUrl = $this->constructFullPath($this->pagename . '/' . $confirmPath);
+			$confirmLink = '<a href="' . $confirmUrl . '">' . __('ml-please-confirm-email') . '</a>';
```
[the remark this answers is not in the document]
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: Updated this troughtout the class

system/datatypes/mailinglist.php
laurensmartina: If you keep it in the same field and the same email address would subscribe, that recor would be re-used.

system/datatypes/mailinglist.php
```diff
@@ -412,246 +439,344 @@ private function unsubscribeAction() {
...
-			notify('error', __('login-to-edit'));
+	private function mailingView(HyphaRequest $request) {
+		$mailingId = $request->getArg(1);
 
+		// check if given id was correct
+		/** @var HyphaDomElement $mailing */
+		$mailing = $this->xml->document()->getElementById($mailingId);
+		if (!$mailing instanceof HyphaDomElement) {
+			notify('error', __('not-found')); // 404
 			return null;
```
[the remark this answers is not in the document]
laurensmartina: updated

system/datatypes/mailinglist.php
laurensmartina: It breaks if I `xpath_encode` the `$code`.

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
matthijskooijman: Is that problematic? And if so, you could also change the subscription code to not reuse an UNSUBSCRIBED status record?

system/datatypes/mailinglist.php
laurensmartina: Yes, it is. :-)

system/datatypes/mailinglist.php
laurensmartina: It was already like this and take too much time to implement, so I consider this outside of the scope of this pull request.

system/datatypes/mailinglist.php
matthijskooijman: Right, I later noticed that sending a message removes it from draft, so that indeed makes sense :-)

system/datatypes/mailinglist.php
matthijskooijman: Why would this take time to implement? I meant this:

```
$commands->append(makeButton(__('send'), 'if(confirm(' . json_encode(__('ml-sure-to-send', array("count" => $num))) . '))' . makeAction($path, self::CMD_SEND, '')));
$commands->append(makeButton(__('ml-test-send'), 'hypha('.json_encode($path).', '.json_encode(self::CMD_SEND_TEST).', prompt(' . json_encode(__('email')) . '), $(this).closest(\'form\'));'));
```

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: fixed it

system/datatypes/mailinglist.php
laurensmartina: I've removed it.

system/datatypes/mailinglist.php
laurensmartina: ... So it is fixed now.. by also replacing `"` by `'`

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: true, but I would rather return in all the WymHTMLForm, it's almost no overhead and way more consistent.

system/datatypes/mailinglist.php
laurensmartina: I blame my laziness.

laurensmartina: fixed it

laurensmartina: thank you for catching this. :-)

system/datatypes/mailinglist.php
laurensmartina: fixed it

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: done

system/datatypes/mailinglist.php
laurensmartina: removed it

system/datatypes/mailinglist.php
laurensmartina: true, but  don't see how that would be an improvement with the current implementation.

laurensmartina: What if I were to move it to an entire other section? That would remove the extra field and prevent any problems.

matthijskooijman: That also seems a bit convoluted. I would just keep it in the same space, to keep the structure simple and equal, and adapt other code to match. Most other code that deals with subscriptions already has to check for status anyway, so that is not particularly hard, right?

We could also consider preserving the current behaviour (removing email) in this PR, and move this change into a separate PR (a separate commit would be a good idea in any case, since it is quite a signficant behaviour change).

laurensmartina: Then we're adding more complexity without any added value. Why is the extra field a problem?

system/datatypes/mailinglist.php
matthijskooijman: Huh? Why is replacing " by ' needed? Javascript strings can be enclosed in " just fine, right? And `json_encode` will have taken care of any needed escaping already? If you replace the quote type, and your encoded strings contain quotes, I think you will actually break things?

system/datatypes/mailinglist.php
matthijskooijman: Because, IMHO, it adds complexity (e.g. the display code must now handle two fields for e-mail rather than one).

laurensmartina: .... I've already retracted my previous comments. sorry. It just didn't click before.. Now it does. :-)

laurensmartina: fixed

matthijskooijman: Yay! Also, deleting comments is sometimes annoying, since I still see them in e-mail notification but cannot use the link in the e-mail anymore to see the comments. Editing them to indicate they are no longr relevant might be better in such cases :-)

system/datatypes/mailinglist.php
```diff
@@ -442,6 +452,77 @@ private function unsubscribeAction(HyphaRequest $request) {
...
+
+		if ($address->getAttribute(self::FIELD_NAME_REMINDED) != true) {
+			$email = $address->getAttribute(self::FIELD_NAME_EMAIL);
+			$address->setAttribute(self::FIELD_NAME_REMINDED, true);
+			$this->xml->saveAndUnlock();
+
+			$this->sendConfirmationMail($code, $email);
+		}
+
+		notify('success', ucfirst(__('ml-reminder-sent')));
```
matthijskooijman: We've discussed this before, whether to show an error or success when the wanted state is already the current state, but in this case I really wonder if this is the right approach. Saying "reminder sent" explicitly indicates that something actually happened. In the case that something is wrong and e-mails are not received, it really does not help to have additional misunderstanding about whether something was even sent in the first place.

I realize that this is probably a corner case because the reminder button will not normally be shown when a reminder was already sent, but users will probably manage to keep open an old interface somewhere and then get confused about the message sent.

system/datatypes/mailinglist.php
```diff
@@ -442,6 +452,77 @@ private function unsubscribeAction(HyphaRequest $request) {
 		return ['redirect', isset($_GET['redirect']) ? $_GET['redirect'] : $this->path()];
 	}
 
+	/**
+	 * @param HyphaRequest $request
+	 * @return null|array
+	 */
+	private function remindAction(HyphaRequest $request) {
+		$code = isset($_GET['code']) ? $_GET['code'] : null;
```
matthijskooijman: Should this authenticate using a code? Why not just check `isUser()`? AFAIU there is never a need for a non-logged-in user to send reminders, so why use a code here? That only leads to additional exposure of the code, which can actually allow (semi-malicious) logged in users to craft a confirmation url, bypassing the usual consent-mechanism for subscriptions.

system/datatypes/mailinglist.php
```diff
@@ -442,6 +452,77 @@ private function unsubscribeAction(HyphaRequest $request) {
...
+		notify('success', ucfirst(__('ml-reminder-sent')));
+		return ['redirect', $this->path(self::PATH_ADDRESSES)];
+	}
+
+	/**
+	 * @param HyphaRequest $request
+	 * @return null|array
+	 */
+	private function deleteAddressAction(HyphaRequest $request) {
+		$code = isset($_GET['code']) ? $_GET['code'] : null;
```
matthijskooijman: Same here: Why use the code rather than logged in status?

system/datatypes/mailinglist.php
matthijskooijman: Reading again I see that the code is also used to identify the e-mail address to remind, but then I would rather pass the e-mailaddress to confirm.

system/datatypes/mailinglist.php
```diff
@@ -442,6 +452,77 @@ private function unsubscribeAction(HyphaRequest $request) {
...
+		if (null == $code) {
+			notify('error', __('missing-arguments'));
+			return null;
+		}
+
+		$this->xml->lockAndReload();
+
+		$xpath = './/' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_CONFIRM_CODE . '=' . xpath_encode($code) . ' and @' . self::FIELD_NAME_STATUS . '!="' . self::ADDRESS_STATUS_UNSUBSCRIBED . '"]';
```
matthijskooijman: This also allows deleting confirmed addresses, which is not what the commit message says. Is that intentional?

system/languages/en.php
```diff
@@ -291,6 +292,11 @@
 		"ml-please-confirm-email" => "Please confirm your subscription by clicking this link",
 		"ml-confirmation-email-subject" => "Please confirm your subscription",
 		"ml-unsubscribe" => "unsubscribe",
+		"ml-send-reminder" => "remind",
+		"ml-reminder-sent" => "reminder sent",
+		"ml-reminded" => "reminded",
```
matthijskooijman: Maybe this should be "already reminded" to emphasize that reminding can only be done once?

system/datatypes/mailinglist.php
```diff
@@ -458,16 +539,38 @@ private function addressesView(HyphaRequest $request) {
...
+						$remindPath = $this->substituteSpecial(self::PATH_REMIND_CODE, ['code' => $confirmCode]);
+						$actions[] = $this->makeActionButton(__('ml-send-reminder'), $remindPath);
+					} else {
+						$statusText .= ', ' . __('ml-reminded');
+					}
+					$deletePath = $this->substituteSpecial(self::PATH_DELETE_ADDRESS_CODE, ['code' => $confirmCode]);
+					$actions[] = $this->makeActionButton(__('delete'), $deletePath);
+				}
+				if ($status === self::ADDRESS_STATUS_CONFIRMED) {
+					$unsubscribeCode = $address->getAttribute(self::FIELD_NAME_UNSUBSCRIBE_CODE);
```
matthijskooijman: Again, I would use an `isAdmin()` check in `unsubscribeAction()` rather than reusing the existing unsubscription code. Here, leaking the code is less of an issue, but checking `isAdmin()` in the action makes it more explicit that unsubscription is also allowed by admins.

system/datatypes/mailinglist.php
laurensmartina: This is indeed an edge case. The remind action can and should only be triggered one time. If the remind action is triggered a second time, there is no need to send another reminder, and also no need to inform the user that this edge case occurred, because for the user the end result is the same.

system/datatypes/mailinglist.php
laurensmartina: fixed it

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: I've updated the code, but still don't think it's needed.

system/languages/en.php
laurensmartina: done

system/datatypes/mailinglist.php
matthijskooijman: > because for the user the end result is the same.

I think this entirely depends on the timeframe. If two users click the remind button at the same time, then the result is pretty much the same, yes: A reminder has been sent. However, if more time passes, then the message might cause a user to (rightfully) assume that *another* reminder has been sent (and e.g. tell the reminded user that he should have received another e-mail, which is then untrue). And even when both presses occur in short succession, this might lead user to believe that *two* reminder emails have been sent, which is not actually the case.

system/datatypes/mailinglist.php
```diff
@@ -442,6 +452,96 @@ private function unsubscribeAction(HyphaRequest $request) {
...
+		$address = $this->findAddresses($xpath)->first();
+
+		if (!$address instanceof HyphaDomElement) {
+			$this->xml->unlock();
+			notify('error', __('invalid-code'));
+			return null;
+		}
+
+		$status = $address->getAttribute(self::FIELD_NAME_STATUS);
+		if ($status === self::ADDRESS_STATUS_CONFIRMED) {
```
matthijskooijman: Would it not be better to reverse this check? E.g. "only pending addresses can be deleted"? Then this code remains working when additional statuses are added later.

system/datatypes/mailinglist.php
matthijskooijman: Really? I do not see any change in `unsubscribeAction` and the url still contains the unsubscribe code?

system/languages/nl.php
```diff
@@ -285,6 +286,13 @@
 		"ml-please-confirm-email" => "Bevestig uw aanmelding door op deze link te klikken.",
 		"ml-confirmation-email-subject" => "Bevestig uw aanmelding",
 		"ml-unsubscribe" => "afmelden",
+		"ml-send-reminder" => "herinner",
+		"ml-reminder-sent" => "herinnering verstuurd",
+		"ml-reminder-already-ent" => "herinnering was al verstuurd",
```
matthijskooijman: typo: ent -> sent

system/datatypes/mailinglist.php
laurensmartina: The unsubscribeAction is used for both admins and subscribers. So the `isAdmin()` isn't a good idea. I could make an extra action for the admins, but why, this one is already there?!

laurensmartina: So, yeah, I left it as it is.

system/datatypes/mailinglist.php
laurensmartina: true

system/datatypes/mailinglist.php
matthijskooijman: That's not really "fixed" :-)

I can imagine that the same action can be authenticated using `isAdmin()`, or failing that, require a code to be specified? Or two actions that call the same function to do the actual work?

As for why: it seems cleaner to me when the authentication code intended to be used for the recipient is not exposed elsewhere. Additionally, adding the isAdmin() check to `unsubscribeAction` makes it more clear to someone reading the code that unsubscription can be done by admins (iow, all access control appears in the same place, namely the action methods).

However, as I said, leaking the unsubscribe code is not as problematic as leaking the confirm code, so when I fail to convince you, leaving this as is is also an option for me :-)

system/languages/nl.php
laurensmartina: fixed

system/datatypes/mailinglist.php
laurensmartina: implemented new method.

laurensmartina:
fixes #269

PR #280, opened 2019-12-16
Language selection

laurensmartina:


system/core/language.php
```diff
@@ -24,9 +24,16 @@ function __($msgid, $args = null) {
...
 	function languageOptionList($select, $omit) {
 		global $isoLangList;
-		$html = '';
-		foreach($isoLangList as $code => $langName) if ($code!=$omit) $html.= '<option value='.$code.($code==$select ? ' selected' : '').'>'.$code.': '.$langName.'</option>';
-		return addslashes($html);
+		$langList = hypha_getUsedContentLanguages();
+
+		$inUse = '';
+		foreach ($langList as $code) if ($code!=$omit) $inUse.= '<option value='.$code.($code==$select ? ' selected' : '').'>'.$code.': '.$isoLangList[$code].'</option>';
+		$html = '<optgroup label="in use">' . addslashes($inUse) . '</optgroup>';
```
matthijskooijman: This is probably a good moment to add html escaping, and remove this pointless addslashes (probably in a subsequent commit, though).

system/core/pages.php
```diff
@@ -100,6 +100,7 @@ function newPage() {
 		html+= '<tr><th><?=__('type')?></th><td><select id="newPageType" name="newPageType">' + '<?php foreach($types as $type => $datatypeName) echo '<option value="'.$type.'"'.($type=='textpage' ? 'selected="selected"' : '').'>'.$datatypeName.'</option>'; ?>' + '</select> ' + infoPageType + '</td></tr>';
 		var infoPageName = <?=json_encode(makeInfoButton('help-page-name'));?>;
 		html+= '<tr><th><?=__('pagename')?></th><td><input type="text" id="newPagename" value="<?=$pagename?>" onblur="validatePagename(this);" onkeyup="validatePagename(this); document.getElementById(\'newPageSubmit\').disabled = this.value ? false : true;"/> ' + infoPageName + '</td></tr>';
+		html+= '<tr><th><?=__('language')?></th><td><select id="newPageLanguage" name="newPageLanguage"><?=languageOptionList('', '')?></select></td></tr>';
```
matthijskooijman: Should this maybe preselect the default or current language? Now, nothing is marked as selected, so it just defaults to the first element in the list.

system/core/pages.php
```diff
@@ -276,18 +277,18 @@ function loadPage(RequestContext $O_O) {
 	registerCommandCallback('newPage', 'newPage');
 	function newPage($newName) {
 		global $hyphaXml, $hyphaUrl, $hyphaContentLanguage;
-
+		$language = array_key_exists('newPageLanguage', $_POST) ? $_POST['newPageLanguage'] : $hyphaContentLanguage;
```
matthijskooijman: I think this should validate the language code, since AFAICS `hypha_addPage` just uses it as-is.

index.php
```diff
@@ -108,6 +108,10 @@
 	 * @deprecated Use O_O instead
 	 */
 	$hyphaLanguage = $O_O->getInterfaceLanguage();
+	$dict = Language::getDictionaryByLanguage($hyphaLanguage);
+	if (array_key_exists('_locale', $dict)) {
```
matthijskooijman: Why not use `__()` here? The only caveat I can think of is that it would return `_locale` if not found, but that seems easy to check?

matthijskooijman:
@laurensmartina I see you pushed to this PR, but it seems my comments are still unfixed. Did you just rebase on top of master maybe?

system/core/language.php
laurensmartina: Tested it, and removing the `addslashes` results in errors.

system/core/pages.php
laurensmartina: fixed

index.php
laurensmartina: fixed

system/core/pages.php
laurensmartina: fixed at `hypha_addPage `

system/core/language.php
matthijskooijman: I've removed the addSlashes and added htmlspecialchars, just force pushed this.

matthijskooijman:
I just rebased this on master and added a few more commits (one at the end, and a handful at the beginning). @laurensmartina, could you review my new commits? If they're ok, I think this is ready to be merged.

system/php-dom-wrapper/Traits/ManipulationTrait.php
```diff
@@ -220,6 +220,8 @@ public function setText($input) {
             // Add new contents in it's place.
             $node->append(new Text($newNodes->getText()));
         });
+
+        return $this;
```
laurensmartina: There is an inconsistency with the return type in the annotation and the actual return type. Please make it consistent.

matthijskooijman:
Thanks for the review. I've fixed the comment, will merge this now.

PR #282, opened 2019-12-29
Adds CSS class to index item indicating public or private state

laurensmartina:
In order to be more flexible in the way the index items can be styled a
CSS class, indicating the public of private state, is added.

### DIFF
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -911,3 +911,7 @@ body.type_peer_reviewed_article .input-wrapper.field_type_editor:not(.field_name
 .share-links .facebook-link{
     background-image: url(images/shareFacebook.png);
 }
+.index-item.is-private:after {
+    content: " *";
+    color: #b90;
+}
--- a/system/core/base.php
+++ b/system/core/base.php
@@ -780,7 +780,7 @@
 			}
 			$privatePos = strpos($pagename, '&#;');
 			if ($privatePos) $pagename = substr($pagename, 0, $privatePos);
-			$htmlList[] = '<div class="index-item type_'.$pageListDatatype[$pagename].'"><a href="'.$language.'/'.$pagename.'">'.showPagename($pagename).'</a>'.asterisk($privatePos).'</div>';
+			$htmlList[] = '<div class="index-item type_'.$pageListDatatype[$pagename].' '.($privatePos ? 'is-private' : 'is-public').'"><a href="'.$language.'/'.$pagename.'">'.showPagename($pagename).'</a></div>';
 		}
 
 		$html = '<div class="index">';


tammoterhark:
I very much like this, but it is not a way for getting rid of the asterisk, which was ment to indicate that a page was not published yet. So I would also like a class indicating the status of the content type. This could be a class that is shown only when the page is not published yet, like .is_not_published

laurensmartina:
> I very much like this, but it is not a way for getting rid of the asterisk, which was ment to indicate that a page was not published yet.

Unfortunately I don't understand what you mean. Do you want to keep the asterisk?

> So I would also like a class indicating the status of the content type. This could be a class that is shown only when the page is not published yet, like .is_not_published

You can accomplish the same with 'is-public' and 'is-private' as you could with 'is_not_published'. Furthermore more consistent to indicate what something is in stead of indication what something is not.

giplt:
Just to be clear, an asterisk doesn't indicate 'not yet published page' but 'private page' as opposed to 'public page'. Recent focus on peer reviewed article content (De Stadsbron) may suggest that the end state of each and any page should be public. However, hypha should facilitate sharing of information within a project group just as well as updating the rest of the world about it. So maybe the difference between the state of a page (private/public) and the state of an article (draft/review/approved/published) is a little confusing here?

tammoterhark:
I guess so.

If what Harmen wrote is true, he mostly is, than I withdraw my comment and
I approve of this PR.

tammoterhark:
yes, but: so what?

An asterisk is a solution, could also be another color of the text, another
font(-weight/-size) etc. These decisions are up to the designer.

Op za 4 jan. 2020 om 19:19 schreef Matthijs Kooijman <
notifications@github.com>:

matthijskooijman:
> An asterisk is a solution, could also be another color of the text, another
font(-weight/-size) etc. These decisions are up to the designer.

Yes, but the default theme should also be usable, so that should somehow indicate private pages as before. Lacking an incentive to change things, I would suggest preserving the asterisk (but through CSS so it can be customized). If we *also* want to change the default design, that's also fine by me, but then we should pick something else. If we want to change the default design to *not* show private pages anymore, then that's also fine, as long as it is an explicit choice (which should then be documented in the commit message at least). I have no real preference here.

laurensmartina:
> > An asterisk is a solution, could also be another color of the text, another
> > font(-weight/-size) etc. These decisions are up to the designer.
> 
> Yes, but the default theme should also be usable, so that should somehow indicate private pages as before. Lacking an incentive to change things, I would suggest preserving the asterisk (but through CSS so it can be customized). If we _also_ want to change the default design, that's also fine by me, but then we should pick something else. If we want to change the default design to _not_ show private pages anymore, then that's also fine, as long as it is an explicit choice (which should then be documented in the commit message at least). I have no real preference here.

Fixed!

PR #288, opened 2020-01-20
Fix quitting the project

matthijskooijman:
This makes quitting the project work again, and fixes the digest message on reincarnating a user.


system/datatypes/settings.php
```diff
@@ -149,14 +149,25 @@ function editQuitMessage() {
...
+				hypha_retireUser($user, /* by_admin */ false);
+				$hyphaXml->saveAndUnlock();
+
+				$subject = __('user-has-left-project', ['name' => $user->getAttribute('fullname'), 'project' => hypha_getTitle()]);
+				$body = nl2br(htmlspecialchars($message));
+				$error = sendMail(getUserEmailList(), $subject, $body);
+				if ($error)
+					notify('error', $error);
+
+				// Leave wil have worked, even when the e-mail failed
```
laurensmartina: typo; will

laurensmartina: type; email

system/core/base.php
```diff
@@ -459,6 +459,24 @@ function hypha_setUser($user, $username, $password, $fullname, $email, $language
 		}
 	}
 
+	/*
+		Function: hypha_retireUser
+		Retires the given user and adds a note to the digest about this.
+	*/
+	function hypha_retireUser($user, $by_admin) {
```
laurensmartina: Passing the user_id in stead of the user object is a bit more clear and less chances on passing the wrong user object, IMO.

system/datatypes/settings.php
matthijskooijman: Fixed.

system/core/base.php
matthijskooijman: Makes sense. Does mean the user must be looked up twice in one of the instances, but that's not really a big deal. Fixed!

matthijskooijman:
Fixed your comments and rebased on top of master, merging now.

PR #290, opened 2020-02-24
peer_reviewed_article: Add featured image

laurensmartina:
peer_reviewed_article: more focus on excerpt
peer_reviewed_article: added feature image

### DIFF
--- a/system/assets/hypha-image-uploader.js
+++ b/system/assets/hypha-image-uploader.js
@@ -1,0 +1,45 @@
+$(function() {
+	$('input[data-image-uploader]').each(function () {
+		// get input
+		var $input = $(this);
+
+		// Add preview
+		var $preview;
+
+		var createPreviewImage = function (src) {
+			if (src) {
+				$preview = $('<img />', {'class': 'preview_uploaded_image', 'src': src}).insertAfter($input);
+				$('<a class="delete-uploaded-image-button" />')
+					.on('click', function (e) {
+						e.preventDefault();
+						// remove delete button
+						$(this).remove();
+
+						// remove image
+						$preview.remove();
+
+						// set as deleted
+						$('#' + $input.data('image-uploader-deleted-field-id')).val(1);
+
+						$input.attr('value', null);
+
+						// toggle input
+						toggleInputAndDelete();
+					})
+					.insertAfter($preview);
+			}
+		};
+
+		var toggleInputAndDelete = function () {
+			$input.toggle(!$input.next().hasClass('preview_uploaded_image'));
+		};
+
+		$input.on('change', function () {
+			createPreviewImage(window.URL.createObjectURL(this.files[0]));
+			toggleInputAndDelete();
+		});
+
+		createPreviewImage($input.data('image-uploader-src'));
+		toggleInputAndDelete();
+	});
+});
--- a/system/datatypes/peer_reviewed_article.php
+++ b/system/datatypes/peer_reviewed_article.php
@@ -56,6 +56,8 @@ class peer_reviewed_article extends HyphaDatatypePage {
 	const FIELD_NAME_AUTHOR = 'author';
 	const FIELD_NAME_STATUS = 'status';
 	const FIELD_NAME_EXCERPT = 'excerpt';
+	const FIELD_NAME_FEATURED_IMAGE = 'featured_image';
+	const FIELD_NAME_FEATURED_IMAGE_DELETED = 'featured_image_deleted';
 	const FIELD_NAME_METHOD = 'method';
 
 	const STATUS_NEWLY_CREATED = 'newly created';
@@ -180,6 +182,7 @@ class peer_reviewed_article extends HyphaDatatypePage {
 				self::FIELD_NAME_CONTEXT => [
 					self::FIELD_NAME_TITLE => [],
 					self::FIELD_NAME_EXCERPT => [],
+					self::FIELD_NAME_FEATURED_IMAGE => [],
 					self::FIELD_NAME_METHOD => [],
 				],
 			],
@@ -418,6 +421,7 @@ class peer_reviewed_article extends HyphaDatatypePage {
 			self::FIELD_NAME_AUTHOR => $this->xml->find(self::FIELD_NAME_ARTICLE)->getAttribute(self::FIELD_NAME_AUTHOR),
 			self::FIELD_NAME_TEXT => $this->xml->find(self::FIELD_NAME_TEXT)->children(),
 			self::FIELD_NAME_EXCERPT => $this->xml->find(self::FIELD_NAME_EXCERPT)->children(),
+			self::FIELD_NAME_FEATURED_IMAGE => $this->xml->documentElement->getOrCreate(self::FIELD_NAME_FEATURED_IMAGE)->text(),
 			self::FIELD_NAME_METHOD => $this->xml->find(self::FIELD_NAME_METHOD)->children(),
 			self::FIELD_NAME_SOURCES => $this->xml->find(self::FIELD_NAME_SOURCES)->children(),
 		];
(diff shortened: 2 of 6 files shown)


system/datatypes/peer_reviewed_article.php
```diff
@@ -437,6 +440,32 @@ public function editAction(HyphaRequest $request) {
...
+				}
+				$orgUrlIndex = $needResize ? 'org' : 'img';
+				if (move_uploaded_file($_FILES[self::FIELD_NAME_FEATURED_IMAGE]['tmp_name'], $destinations[$orgUrlIndex])) {
+					if ($needResize) {
+						image_resize($destinations[$orgUrlIndex], $destinations['img'], $maxSize[0], $maxSize[1]);
+					}
+				}
+
+				$newImage = $destinations['img'];
+			}
```
matthijskooijman: This silently ignores images that are too big or have an invalid extension.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1048,6 +1078,11 @@ private function createEditForm(array $values = []) {
 					<strong><label for="[[field-name-excerpt]]">[[excerpt]]</label></strong>[[info-button-excerpt]]<br><editor name="[[field-name-excerpt]]"></editor>
 				</div>
 			</div>
+			<div class="section" style="padding:5px; margin-bottom:5px; position:relative;">
+				<div class="input-wrapper field_type_editor field_name_[[field-name-excerpt]]">
```
matthijskooijman: This should be featured-image rather than excerpt?

system/assets/hypha-image-uploader.js
```diff
@@ -0,0 +1,55 @@
...
+		// Add preview
+		var $preview;
+
+		$input.on('change', function () {
+			createPreviewImage(window.URL.createObjectURL(this.files[0]));
+			toggleInputAndDelete();
+		});
+		var createPreviewImage = function (src) {
+			if (src) {
+				$preview = $('<img class="preview_uploaded_image" src="' + src + '">').insertAfter($input);
```
matthijskooijman: This probably needs some htmlencoding for the src attribute (probably easiest to generate an img with src and then use the jquery attribute set method to set the src).

system/assets/hypha-image-uploader.js
```diff
@@ -0,0 +1,55 @@
...
+				$('<a href="#"><img src="https://image.flaticon.com/icons/svg/261/261935.svg" style="width:40px;height:40px;" accept="image/png,image/jpeg,image/jpg,image/gif,image/bmp"></a>')
```
matthijskooijman: What's the point of the `accept` attribute here? I could not find any info on it?

Also, this link and/or img should probably get a CSS class and the width/height should be in hypha.css.

system/assets/hypha-image-uploader.js
```diff
@@ -0,0 +1,55 @@
...
+						// call delete function
+						if ($input.prop('defaultValue') == $preview.attr('src')) {
+							$.post('delete_uploaded_image', {src: $preview.attr('src')});
+						}
+
+						// remove delete button
+						$(this).remove();
+
+						// set image src to null
+						$preview.attr('src', null);
```
matthijskooijman: Why bother with this when you're about to remove the preview entirely?

system/assets/hypha-image-uploader.js
```diff
@@ -0,0 +1,55 @@
+$(function() {
+	$('[data-image-uploader]').each(function () {
+		// get input
+		var $input = $(this);
+		console.log($input.data('image-uploader-src'));
+
+		var src = $input.data('image-uploader-src');
```
matthijskooijman: Maybe this line should be moved way further down, since the `src` var is only used there (and now it might be confusing with the `src` parameter to `createPreviewImage()`.

system/core/pages.php
```diff
@@ -194,6 +194,12 @@ function loadPage(RequestContext $O_O) {
 				$helpLanguage = isset($args[1]) ? $args[1] : $O_O->getInterfaceLanguage();
 				echo hypha_searchHelp($O_O, $subject, $helpLanguage);
 				exit;
+			case HyphaRequest::HYPHA_SYSTEM_PAGE_DELETE_UPLOADED_IMAGE:
+				$src = $O_O->getRequest()->getPostValue('src');
+				$success = unlink($src);
```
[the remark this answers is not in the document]
matthijskooijman: Also, this only deletes the resized image, not the original (if it exists).

system/assets/hypha-image-uploader.js
```diff
@@ -0,0 +1,55 @@
+$(function() {
+	$('[data-image-uploader]').each(function () {
+		// get input
+		var $input = $(this);
+		console.log($input.data('image-uploader-src'));
```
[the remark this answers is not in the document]
laurensmartina: fixed

system/core/pages.php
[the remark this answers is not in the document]
laurensmartina: oopsie. good point!

system/datatypes/peer_reviewed_article.php
tammoterhark: Please: no inline styling...

system/core/pages.php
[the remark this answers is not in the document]
laurensmartina: fixed

system/datatypes/peer_reviewed_article.php
laurensmartina: yes

laurensmartina: @tammoterhark : is all over the file.. I consider this to be out of the scope of this pull request.

system/assets/hypha-image-uploader.js
laurensmartina: Seems logical, but actually this is needed to work properly.

system/assets/hypha-image-uploader.js
laurensmartina: placed the accept at the input field, where it belongs.

system/assets/hypha-image-uploader.js
laurensmartina: fixed

system/assets/hypha-image-uploader.js
laurensmartina: moved styling to css

system/assets/hypha-image-uploader.js
laurensmartina: found a way to fix it

system/assets/hypha-image-uploader.js
laurensmartina: fixed

laurensmartina:
> Overall, I wonder about the duplication in code between the article edit form handling and the wymeditor upload handling. Can we somehow unify these two?
> 
> I can imagine that the editor does a background upload (like the wymeditor) and then puts the resulting url/filename in the edit form, rather than uploading the image as part of the edit form itself.
> 
> Alternatively, I can imagine that the code is slightly generalized and can be called from both the wymeditor upload handler as well as from the article edit form handler? Or maybe the HTMLForm class could support this somehow?

I agree with all of you comments. And would like to fix this later on.

tammoterhark:
Image upload to featured image seems to work at first, but after saving the article and re-opening, the image has disappeared.

When I look on disk, the image is there, but it just does not render. Showed to to Laurens.

system/datatypes/peer_reviewed_article.php
laurensmartina: fixed

system/datatypes/peer_reviewed_article.php
```diff
@@ -416,14 +429,49 @@ public function editAction(HyphaRequest $request) {
...
+
+		$featuredImageData = $form->dataFor(self::FIELD_NAME_FEATURED_IMAGE);
+		$orgImg = $this->xml->documentElement->getOrCreate(self::FIELD_NAME_FEATURED_IMAGE)->getText();
+		$hasNewImg = $featuredImageData && !$featuredImageData['error'];
+
+		if ($hasNewImg) {
+			$ext = strtolower(substr(strrchr($featuredImageData['name'], '.'), 1));
+			@$size = getimagesize($_FILES[self::FIELD_NAME_FEATURED_IMAGE]['tmp_name']);
+			if (!$size || !in_array($ext, ['jpg','jpeg','png','gif','bmp'])) {
+				$form->errors[self::FIELD_NAME_FEATURED_IMAGE] = __('invalid-image-file').ini_get('upload_max_filesize');
```
matthijskooijman: Why append the upload max filesize, why not use the interpolation support that __ now has?

system/datatypes/peer_reviewed_article.php
```diff
@@ -416,14 +429,49 @@ public function editAction(HyphaRequest $request) {
...
 
+		$formData = $request->getPostData() + $_FILES;
+
 		// create form
-		$form = $this->createEditForm($request->getPostData());
+		$form = $this->createEditForm($formData);
+
+		$featuredImageData = $form->dataFor(self::FIELD_NAME_FEATURED_IMAGE);
+		$orgImg = $this->xml->documentElement->getOrCreate(self::FIELD_NAME_FEATURED_IMAGE)->getText();
+		$hasNewImg = $featuredImageData && !$featuredImageData['error'];
```
matthijskooijman: If `$featuredImageData['error']` is set, then the upload is completely ignored, without a message to the user. Is that ok? When can this happen? Doesn't this include "File too big"?

system/datatypes/peer_reviewed_article.php
```diff
@@ -416,14 +429,49 @@ public function editAction(HyphaRequest $request) {
...
+
+		if ($hasNewImg) {
+			$ext = strtolower(substr(strrchr($featuredImageData['name'], '.'), 1));
```
matthijskooijman: Maybe use `pathinfo` here instead?

system/datatypes/peer_reviewed_article.php
```diff
@@ -416,14 +429,49 @@ public function editAction(HyphaRequest $request) {
...
+			$destinations = ['org' => 'data/images/org/' . $filename, 'img' => 'data/images/' . $filename];
+			if ($needResize && !file_exists('data/images/org/')) {
+				mkdir('data/images/org/', 0777, true);
+			}
+			$orgUrlIndex = $needResize ? 'org' : 'img';
+			if (move_uploaded_file($_FILES[self::FIELD_NAME_FEATURED_IMAGE]['tmp_name'], $destinations[$orgUrlIndex])) {
+				if ($needResize) {
+					image_resize($destinations[$orgUrlIndex], $destinations['img'], $maxSize[0], $maxSize[1]);
+				}
+			}
```
matthijskooijman: This still ignores a failure of `move_uploaded_file` or `image_resize` (though that might trigger a PHP error, of course, which might be sufficien).

system/datatypes/peer_reviewed_article.php
```diff
@@ -434,6 +482,13 @@ public function editAction(HyphaRequest $request) {
 		$this->xml->find(self::FIELD_NAME_EXCERPT)->setHtml($form->dataFor(self::FIELD_NAME_EXCERPT));
 		$this->xml->find(self::FIELD_NAME_METHOD)->setHtml($form->dataFor(self::FIELD_NAME_METHOD));
 		$this->xml->find(self::FIELD_NAME_SOURCES)->setHtml($form->dataFor(self::FIELD_NAME_SOURCES));
+		if ($newImage) {
+			$this->xml->documentElement->getOrCreate(self::FIELD_NAME_FEATURED_IMAGE)->setText($newImage);
+		}
+
+		if ($orgImg !== '' && $form->dataFor(self::FIELD_NAME_FEATURED_IMAGE_DELETED)) {
+			unlink($orgImg);
```
matthijskooijman: Why wait until here to unlink? If you unlink in the if above, you probably do not need the `$orgImg` variable at all (since before setting a new image you can look at the original value in the XML).

Also, this still unlinks only the resized version, not the original (if present).

system/datatypes/peer_reviewed_article.php
laurensmartina: is also like this in `loadPage()` in `pages.php` (`case HyphaRequest::HYPHA_SYSTEM_PAGE_UPLOAD`). this needs to be centralised and improved.

matthijskooijman:
I've copied the excerpt commit to #299, leaving the featured image commits here for more discussion and improvement. I also rebased this branch on top of #299, which is rebased on top of current master.

PR #294, opened 2020-03-30
tags: Sort tag lists on display

matthijskooijman:
This uses natural (e.g 2 before 10) and case insensitive sorting, since
that is probably most natural for users.


matthijskooijman:
This has been running on destadsbron.nl for a few weeks, so I'm just going to merge this.

PR #300, opened 2020-04-22
JS: Import the used jquery version locally

matthijskooijman:
For some reason, code.jquery is often very slow, leading to a large
latency loading the page. Better just host our own jquery, so hypha is
now self-contained (except for external links to licensing images in the
default footer, but those are not essential).


matthijskooijman:
This is simple enough, so I'm merging this without further review.

Issue #298, opened 2020-04-08
Minimal tag-based index


matthijskooijman:
For rubrics at de stadsbron, a minimal version would be:
 - A fixed system page like the current index page
 - Which shows articles with one tag, which is specified through an URL parameter
 - Articles are displayed using a title and excerpt (needs #290), no image yet.
 - This page can then be linked from the tags displayed on any other page
 - This should be sorted by publication date, which is done by loading the page XML for each page displayed (this is inefficient, but removes the need for migrating these dates to the index initially).

Some code for this is probably already present in the https://github.com/PlanBCode/hypha/tree/datatype-tag_list branch (of which this is a slimmed down version).

Later improvements:
 - Include a featured image for articles
 - Include text pages and other datatypes
 - Put a sort date in the index (copy of the publication date for articles) to allow sorting on that (requires running a migration for existing articles).
 - Allow visitors to  customize sorting behaviour.
 - Create a new datatype to show such an index, to allow adding a header, footer, custom sorting, etc. This is #127.

matthijskooijman:
For now, the plan is to maybe *not* extend the current tag page (much) further, but instead implement #87 to generate homepages and tag-based "dossier" pages. The code that generates the lists for #87 should probably be based on the current tag index, though, and could maybe be reused by the current tag index to generate the actual content.

Also, for now the idea is to just create these "dossier" pages separately, without linking the tags on a page to these pages (they would still go to the current tag index system page). On the longer term, we could maybe add (manual) per-tag configuration for this (e.g. "when you click this tag, go to <this> (text)page").

PR #301, opened 2020-04-22
Tag index page

matthijskooijman:
This implements a minimal tag index page (i.e. #298). It is a system page, which shows all pages (not just articles) tagged with a given tag. These index pages can be reached by clicking a tag on an article.

### DIFF
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -92,6 +92,17 @@
 			return showPagename($this->pagename);
 		}
 
+		public function renderSingleLine(HyphaDomElement $container) {
+			$h2 = $container->document()->createElement('h2');
+			$h2->setText($this->getTitle());
+			$h2->appendTo($container);
+		}
+
+		public function renderExcerpt(HyphaDomElement $container) {
+			// Default to just the single line version
+			$this->renderSingleLine($container);
+		}
+
 		/*
 		 * Returns a date typically used for sorting these pages
 		 * (i.e. the publish date or last update date).
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -81,6 +81,15 @@
 			$this->language = $language->getAttribute('id');
 			$this->pagename = $language->getAttribute('name');
 		}
+
+		/*
+		 * Returns a date typically used for sorting these pages
+		 * (i.e. the publish date or last update date).
+		 *
+		 * Returns a DateTime object or null if no timestamp is
+		 * available.
+		 */
+		abstract public function getSortDateTime();
 	}
 
 	/*
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -181,6 +181,21 @@ EOF;
 	}
 
 	/*
+		Function: createPageInstance
+		Creates a page object (appropriate subclass of
+		HyphaDatatypePage, e.g. textpage) instance from the
+		given page node.
+
+		Parameters:
+		RequestContext $O_O
+		HyphaDomElement $node - The page node from $hyphaXml
+	*/
+	function createPageInstance(RequestContext $O_O, HyphaDomElement $node) {
+		$type = $node->getAttribute('type');
+		return new $type($node, $O_O);
+	}
+
+	/*
 		Function: loadPage
 		loads all html needed for page pagename/view
 
@@ -207,8 +222,7 @@ EOF;
 
 			$isPrivate = $_node && in_array($_node->getAttribute('private'), ['true', '1', 'on']);
 			if ($_node && (!$isPrivate || isUser())) {
-				$_type = $_node->getAttribute('type');
-				$hyphaPage = new $_type($_node, $O_O);
+				$hyphaPage = createPageInstance($O_O, $_node);
 
 				// add tag list to all non-system pages
 				$hyphaHtml->writeToElement('tagList', hypha_indexTags($_node, $hyphaPage->language));
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -305,6 +305,9 @@ EOF;
 						break;
 				}
 				break;
+			case HyphaRequest::HYPHA_SYSTEM_PAGE_TAG_INDEX:
+				$hyphaPage = new tagindexpage($O_O);
+				break;
 			case HyphaRequest::HYPHA_SYSTEM_PAGE_HELP:
 				$subject = isset($args[0]) ? urldecode($args[0]) : 'undefined';
 				$helpLanguage = isset($args[1]) ? $args[1] : $O_O->getInterfaceLanguage();
--- a/system/core/pages.php
+++ b/system/core/pages.php
@@ -4,23 +4,54 @@
 	/*
 		Title: Pages
 
-		This chapter describes the way pages are registered.
+		This chapter describes the way views and pages are registered.
 	*/
 
 	/*
-		Class: Page
-		abstract class for handling a certain kind of data
+		Class: HyphaPage
+		abstract class for a page that can process a request.
 	*/
-	abstract class Page {
-		public $pageListNode, $html, $language, $pagename, $O_O, $args, $privateFlag;
-		function __construct($node, RequestContext $O_O) {
+	abstract class HyphaPage {
+		public $html, $language, $pagename, $O_O, $args, $privateFlag;
+
+		function __construct(RequestContext $O_O) {
 			global $hyphaHtml;
 			$this->html = $hyphaHtml;
 			$this->O_O = $O_O;
 			// TODO: Remove args (and let subclasses talk to the request instead)
 			$this->args = $O_O->getRequest()->getArgs();
-			if ($node)
-				$this->replacePageListNode($node);
+		}
+
+		/**
+		 * Return the given url argument, or null if it is not
+		 * present.
+		 */
+		protected function getArg($index) {
+			if (array_key_exists($index, $this->args))
+				return $this->args[$index];
+			return null;
+		}
+
+		abstract function process(HyphaRequest $request);
+	}
+
+	/*
+		Class: HyphaSystemPage
+		abstract class for a system page that has no backing datatype.
+	*/
+	abstract class HyphaSystemPage extends HyphaPage {
+	}
+
+	/*
+		Class: HyphaDatatypePage
+		abstract class for a page backed by datatype and xml file.
+	*/
+	abstract class HyphaDatatypePage extends HyphaPage {
+		public $pageListNode, $language, $pagename, $privateFlag;
+
+		function __construct($node, RequestContext $O_O) {
+			parent::__construct($O_O);
+			$this->replacePageListNode($node);
 		}
 
 		public static function getDatatypeName() {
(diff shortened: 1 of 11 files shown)


system/core/pages.php
```diff
@@ -4,21 +4,53 @@
...
 			// TODO: Remove args (and let subclasses talk to the request instead)
 			$this->args = $O_O->getRequest()->getArgs();
+		}
+
+		/**
+		 * Return the given url argument, or null if it is not
+		 * present.
+		 */
+		protected function getArg($index) {
+			if (array_key_exists($index, $this->args))
```
RoukePouw, line 30: I would suggest using ternary operator (as a non blocking review suggestion :-) )

system/core/pages.php
```diff
@@ -4,21 +4,53 @@
...
+		abstract class for a page that can process a request.
 	*/
-	abstract class Page {
-		public $pageListNode, $html, $language, $pagename, $O_O, $args, $privateFlag;
-		function __construct($node, RequestContext $O_O) {
+	abstract class HyphaPage {
+		public $html, $language, $pagename, $O_O, $args, $privateFlag;
+
+		function __construct(RequestContext $O_O) {
 			global $hyphaHtml;
```
RoukePouw, line 18: Curious if this global could not be passed through constructor? (as a non blocking review suggestion :-) )

system/core/pages.php
```diff
@@ -4,21 +4,53 @@
...
+	}
+
+	/*
+		Class: HyphaDatatypePage
+		abstract class for a page backed by datatype and xml file.
+	*/
+	abstract class HyphaDatatypePage extends HyphaPage {
+		public $pageListNode, $language, $pagename, $privateFlag;
+
+		function __construct($node, RequestContext $O_O) {
```
RoukePouw, line 52: if `$node` is optional, perhaps add it as second, optional parameter?

system/core/pages.php
```diff
@@ -51,17 +83,34 @@ protected function replacePageListNode($node) {
...
-			return null;
+		public function getTitle() {
+			// TODO: Implement a meaningful title in all
+			// subclasses and make this function abstract?
+			return showPagename($this->pagename);
 		}
 
-		abstract function process(HyphaRequest $request);
+		public function renderSingleLine($container) {
+			$h2 = $container->document()->createElement('h2');
```
RoukePouw, line 97: Not sure how DOM is handled but I've learned the best practice is to first create the element and then add it to the DOM (as a non blocking review suggestion :-) )

system/core/pages.php
```diff
@@ -161,6 +210,21 @@ function newPage() {
...
+		Function: createPageInstance
+		Creates a page object (appropriate subclass of
+		HyphaDatatypePage, e.g. textpage) instance from the
+		given page node.
+
+		Parameters:
+		RequestContext $O_O
+		HyphaDomElement $node - The page node from $hyphaXml
+	*/
+	function createPageInstance(RequestContext $O_O, HyphaDomElement $node) {
```
RoukePouw, line 223: I observe TypeDefs (which I like). Should we add these to all functions or do we require lower PHP version support?

system/core/pages.php
```diff
@@ -242,6 +305,9 @@ function loadPage(RequestContext $O_O) {
 						break;
 				}
 				break;
+			case HyphaRequest::HYPHA_SYSTEM_PAGE_TAG_INDEX:
+				$hyphaPage = new tagindexpage($O_O);
```
RoukePouw, line 309: I notice that this function is not camel cased. (as a very minor non blocking review suggestion :-) )

system/core/tags.php
```diff
@@ -3,6 +3,64 @@
 	// - implement error notification if new tag exists
 	// - implement tag management page: edit, translate, remove or merge tags
 	// - implement tagcontainer pagetype
+	class HyphaTags {
+		static function findTagByLabel($lang, $label) {
+			/** @var HyphaDomElement $hyphaXml */
+			global $hyphaXml;
+			$path = 'hypha/tagList/tag/language[@id=' . xpath_encode($lang) . ' and @label=' . xpath_encode($label) . ']';
+			$node = $hyphaXml->findXPath($path)->parent()->first();
+			if (!$node)
```
RoukePouw, line 12: Would suggest a ternary (as a non blocking review suggestion :-) )

system/core/tags.php
```diff
@@ -3,6 +3,64 @@
...
+				return null;
+			return new HyphaTag($node);
+		}
+
+		static function findPagesWithTags(HyphaTag $tag, array $pagetypes=[], bool $includePrivate=false, $skip=0, $limit=0) {
```
RoukePouw, line 17: Yay, more typedefs!

system/core/tags.php
```diff
@@ -3,6 +3,64 @@
...
+			/** @var HyphaDomElement $hyphaXml */
+			global $hyphaXml;
+
+			$pagefilters = '';
+			if ($pagetypes) {
+				$typefilter = function($type) { return "@type=" . xpath_encode($type); };
+				$attr_filters = array_map($typefilter, $pagetypes);
+				$pagefilters .= "[" . implode(" or ", $attr_filters) . "]";
```
RoukePouw, line 25: Varied non camel, underscore variable naming (as a non blocking review suggestion :-) )

system/datatypes/festival.php
```diff
@@ -97,6 +97,11 @@ function process(HyphaRequest $request) {
 			}
 		}
 
+		public function getSortDateTime() {
+			// TODO: What to return here?
```
RoukePouw, line 101: Is this WIP or left for a later time/issue?

system/datatypes/mailinglist.php
```diff
@@ -125,6 +125,11 @@ public function process(HyphaRequest $request) {
 		return '404';
 	}
 
+	public function getSortDateTime() {
+		// TODO: Return last mailing sent date?
```
RoukePouw, line 129: Is this WIP or left for a later time/issue?

system/datatypes/peer_reviewed_article.php
```diff
@@ -139,6 +139,25 @@ public function process(HyphaRequest $request) {
...
+		$timestamp = ltrim($value, "t");
+		return new DateTime("@" . $timestamp);
+	}
+
+	public function renderExcerpt($container) {
+		$doc = $container->document();
+		$doc->createElement('h2')->appendTo($container)->setText($this->getTitle());
+		$article = $this->xml->find(self::FIELD_NAME_ARTICLE);
+		$this->appendAuthorAndTimestamp($container, $article);
+		$excerpt_div = $doc->createElement('div')->appendTo($container);
```
RoukePouw, line 156: Same comment as before. I would suggest first building the element and then add it to the DOM

system/datatypes/tagindex.php
```diff
@@ -0,0 +1,94 @@
...
+				// TODO Can we better handle null here?
+				return ($b_date ? $b_date->getTimestamp() : 0) - ($a_date ? $a_date->getTimestamp() : 0);
+			};
+			usort($pages, $compare);
+			return $pages;
+		}
+
+		private function getPages(HyphaTag $tag, $includePrivate = false) {
+			$pagetypes = ['peer_reviewed_article'];
+			$pagetypes = [];
```
RoukePouw, line 61: variable is overwritten here

system/datatypes/tagindex.php
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		private function renderPage(HyphaDatatypePage $page) {
+			$li = $this->html->createElement('li');
+			$li->addClass('tagindex-item');
+			$li->addClass('type_' . get_class($page));
+			$li->addClass($page->privateFlag ? 'is-private' : 'is-public');
+
+			$a = $this->html->createElement('a')->appendTo($li);
+			$a->setAttribute('href', $page->language.'/'.$page->pagename);
```
RoukePouw, line 88: are special characters handled here? (double checking because I read something in the chat)

system/datatypes/mailinglist.php
matthijskooijman: Left for later indeed.

system/datatypes/festival.php
matthijskooijman: Left for later indeed.

system/core/pages.php
matthijskooijman: Yeah, we started adding type annotations on arguments at some point, so ideally we add them on all functions we add or change. Maybe an overall refactor to add them everywhere might be good. but also some work.

system/core/pages.php
matthijskooijman: The datatype class names use lowercase naming. Maybe we should change that at some point, but for now I'm just staying consistent with the existing classes.

system/datatypes/tagindex.php
matthijskooijman: This leaves urlencoding of special characters (in particular spaces) to the browser, at some point url generation should really be generalized and do urlencoding explicitly. Special characters are handled at the HTML level properly by DOMDocument here.

system/core/pages.php
matthijskooijman: Yeah, ideally we would get rid of all globals at some point, but that would be a bigger refactor than I want to do now.

RoukePouw: Understood :-)

system/core/pages.php
matthijskooijman: Since I only moved this code in this PR, I'm inclined to leave this as-is. Also, I'm not quite convinced that this really is more readable with at ternary if. With the PHP7 `??` operator, it would become more compact (and also subtly different if `null` values could be present in the array), but we still aim for PHP5.6 support.

system/core/pages.php
matthijskooijman: Actually, `$node` is no longer optional, since the argument was moved down from `Page` into `HyphaDataPage` where we can require it. This does mean the `if ($node)` is no longer required, so I'll remove that one :-)

system/core/pages.php
matthijskooijman: I suspect that in a browser, this has performance advantages (not changing something that is already being rendered), but I also suspect it does not matter much in this case, since the DOM is only kept in memory. Still, it is probably good to stay consistent, so I've changed this.

system/core/tags.php
matthijskooijman: Here I think it would indeed help readability, so I've made this a ternary if.

system/core/tags.php
matthijskooijman: Good point. My personal preference is often lowercase underscore naming for variables, so somethings I forget using the proper style. Thanks, fixed now.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Fixed.

system/datatypes/tagindex.php
matthijskooijman: Ah, that was old code, I removed the first line.

matthijskooijman:
I've force-pushed some more fixes based on @RoukePouw's review (fixup'd into the original commits). Ready to merge as far as I'm concerned, after some more testing by @tammoterhark and/or @dianawi.

PR #307, opened 2020-05-07
Fix regex, add comments

### LINKED ISSUE #306: Install new hypha fails (probably because of empty lang dropdown)
when creating a hypha.php to download the subsequent form has an empty language dropdown. On send a blank screen is shown and the install did not succeed

RoukePouw:
Fixes #306

### DIFF
--- a/hypha.php
+++ b/hypha.php
@@ -117,7 +117,7 @@
 		$data .= "			//END_OF_DATA\n";
 
 		// include data library
-		$hypha = preg_replace('#^\t*//START_OF_DATA\n.*//END_OF_DATA\n#ms', $data, $hypha);
+		$hypha = preg_replace('#\t*//START_OF_DATA\n.*//END_OF_DATA\n#ms', $data, $hypha);
 
 		// push script to client
 		header('Content-Type: application/octet-stream');
@@ -415,13 +415,16 @@
 		<build>
 	*/
 	function data($name) {
+	    $zip = false;
 		switch ($name) {
+		    // Note: this switch data will be overwritten with file content data for the distributable
 			//START_OF_DATA
 			case 'index': return '';
 			case 'system/languages/languages.json': return file_get_contents('system/languages/languages.json');
 			//END_OF_DATA
 		}
-		if ($zip) return gzdecode(base64_decode($zip));
+		if ($zip) return gzdecode(base64_decode($zip)); // $zip can be set by the contents that will be written to the distributable
+        //TODO do we need to throw of return something here?
 	}
 
 	session_start();


matthijskooijman:
Hm, I'm a bit surprised about this, since:
 - This regex was introduced like this in a876c60 and I assume I would have tested this then (since the entire point of that commit was to fix hypha.php building).
 - The "m" flag in the regex should make `^` match the start of each line, not just the start of the string. E.g. from [the perl docs](https://perldoc.perl.org/perlre.html#Modifiers):
   > `m`
   > Treat the string being matched against as multiple lines. That is, change "^" and "$" from matching the start of the string's first line and the end of its last line to matching the start and end of each line within the string.

Even more, I just tested this here (PHP 7.3 on Ubuntu) and it seems it just works. If I go to hypha.php, fill in login details and click build, I get a hypha.php *with* data filled in. i.e.:

```
			//START_OF_DATA
			case 'index': $zip = "H4sIAAAAAAACA41U7W7CMAx8IWifAbFOTIJpYkj7iUJqICyJszih9O2X8k3dsv2L7k7O+WJH2RIOmdu6QbYNQkogGlBNAUwu0UNeoowGbDhK7om98AojMXyFFYFncEwgF8M+leawExvoEJcqIK88qd1WzOEnAnGXK0HAwCA2vLr0tQvI4FIE0VnkqzaTxWz6it4w7uxmjDbAgZvSwm5iapF7QGOiVVIEhZaxRRPWwgv1ULExGGoHlK/TlWovdDdLEIKyj43fWCOUTqxW1FM8Zaauo8LZVps3xgH4pYe9ggrKpfBBSQ3d0uy+gYyqq6Sqzentj8lxdJfS9nV2BTKjbLYjLlRrLwxwnL6V7ZA7HTedRMZu3FG3YYcuurRZRv/D9s1yCmBYohlWXjgHPn85b+EiBXUfXls3Rq1BNtPTW2rReqo2P9L6Gf2OJUxbU9LWzIvRy6zoYwsN7Q+lr91nmuMi9Hb54bH5ytJIv1kKPsr2RvHkzF83jok+oYkX/UVy2WTK7UNqN/x6Sq+LlvNwcnX5Zk+bHw7hF2MY1iqYBQAA"; break;
(... more data ...)
```

Are you sure this isn't working for you and the regex is the problem? Or could there some other factor that just coincides with your change in the regex?

RoukePouw:
I'm tried a clean install again and now it did work :-S I'm thoroughly lost what went wrong yesterday. 

In my download folder I still have 27kb hypha.php files that do not have the file data replaced, downloaded after I did a clean install yesterday. 

I then altered to code echo instead of download the file to directly see what was happening, when I changed the regex I finally saw the data in the output. I then re-enabled the download and ended up with correct 695kb files.

matthijskooijman:
Hm, something weird is going on. However, that makes it likely that the regex was not the actual problem, and this PR will not really fix anything.

I'm inclined to close this PR, for that reason.

If you can manage to reproduce the problem, we can try to figure out what is wrong and fix it (in #306), but if not, I'm also inclined to just close that issue.

RoukePouw:
> I'm inclined to close this PR, for that reason. [...] I'm also inclined to just close that issue.

Agreed :-) If I stumble upon it happening again I'll reopen it.

PR #308, opened 2020-05-08
Improve feedback on invalid image upload

RoukePouw:


matthijskooijman:
Looks good at first glance. However, I do wonder: Where does the upload form get handled normally? Isn't there some error handling there already?

I believe that the upload plugin we're using now is (mostly) unmodified from the wymeditor version, we just plugged in a custom backend to handle the uploads. So maybe there is already working error handling, and a better (but not more complicated) fix would be to change the backend to return errors in a way the frontend code expects?

RoukePouw:
this gets passed `null` while it seems to expect an array. Will need to look further to see why it ends up with null here and not with backend response `<script>...uploadReponse('~error')...`

```
jquery.wymeditor.image-upload.js:41
complete: function(response){
				response = response[0];
				jQuery(".wym_src", doc).val(response.thumbUrl);
				jQuery(".wym_alt", doc).val(response.original_filename);
				jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);
			}
```

RoukePouw:
This is the response  on a good upload: 

```
 [{original_filename: "quest.png", downloadUrl: "http://localhost:8888/images/5eb53066f3abe.png", thumbUrl: "http://localhost:8888/images/5eb53066f3abe.png"}] 
```

That comes from 

```
pages.php:360
if ($return !== null) {
						echo json_encode($return);
					} else {
					    echo '<script language="javascript" type="text/javascript">window.top.window.uploadResponse(\''.$response.'\');</script>';
					}
```

I can replace the script with json but I haven't found the attribute to indicate an error (perhaps through https statuscode? (How can I make hypha return a 500?)

matthijskooijman:
> this gets passed null while it seems to expect an array. Will need to look further to see why it ends up with null here and not with backend response 

I suspect it gets passed null, because the string is invalid json (and jquery [is told to expect JSON](https://github.com/PlanBCode/hypha/blob/c04fc82159b0c3355d72029a72ac876e93d5805b/system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js#L37)).

That looks like returning the error message in JSON is actually a good idea, then the success and error response are both JSON which is nice and consistent. However, it *does* seem that the original plugin code has no way to handle errors, which is a bit sad...

So, I would suggest we just add the error handling (probably just with an alert as before) in the image upload plugin, e.g. [in here](https://github.com/PlanBCode/hypha/blob/c04fc82159b0c3355d72029a72ac876e93d5805b/system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js#L41-L46). It would be elegant to use a HTTP status code and have some kind of error handler, but it actually seems that the ["iframe post form plugin"](https://github.com/PlanBCode/hypha/blob/master/system/wymeditor/plugins/jquery.iframe-post-form.js) does not really support this (not sure if it could, using this iframe approach). So, just returning json with an `errror` field and a message, check for that, and do an alert if `error` is present is probably a clean approach?

(I am wondering if we could maybe ditch the entire "iframe" approach somehow with newer web APIs or jquery plugins, but maybe we should not make this bigger than needed for now...)

RoukePouw:
> So, I would suggest we just add the error handling (probably just with an alert as before) in the image upload plugin, 

Done :-) 

Can we now also remove the ~ in front of the error messages or is that this used someplace else?

>  but maybe we should not make this bigger than needed for now...

 I'm very much in favor of baby steps ;-) Though I fully share the view that lossing the iframe would be a good thing :-)

system/core/pages.php
```diff
@@ -360,7 +360,7 @@ function loadPage(RequestContext $O_O) {
 					if ($return !== null) {
 						echo json_encode($return);
 					} else {
-						echo '<script language="javascript" type="text/javascript">window.top.window.uploadResponse(\''.$response.'\');</script>';
+						echo '[{"error":"'.$response.'"}]';
```
matthijskooijman, line 363: This could run into problems with escaping (e.g. when there is a `"` in `$response`). I would suggest:

    if ($return === null)
        $return = [['error' => $response]];

And then let the existing `json_encode` handle (dropping the if before it).

There is some additional refactoring that could maybe be done here, but some of this code might be replaced by Laurens' backend changes anyway.

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -40,9 +40,14 @@ WYMeditor.editor.prototype.image_upload = function() {
 			},
 			complete: function(response){
 				response = response[0];
-				jQuery(".wym_src", doc).val(response.thumbUrl);
-				jQuery(".wym_alt", doc).val(response.original_filename);
-				jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);
+				if(response.error){
+                    alert(response.error.substring(1));
+                    document.getElementById('upload-image').value = oldSubmitLabel;
```
matthijskooijman, line 45: On error, you now do

    document.getElementById('upload-image').value = oldSubmitLabel;

and on success

                    jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);

Those look like they might be doing the same, so maybe this can be one line outside of the if? Of is there a difference?

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -40,9 +40,14 @@ WYMeditor.editor.prototype.image_upload = function() {
...
+                }else {
+                    jQuery(".wym_src", doc).val(response.thumbUrl);
+                    jQuery(".wym_alt", doc).val(response.original_filename);
+                    jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);
+                }
```
matthijskooijman, line 50: jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
RoukePouw: Ah, missed that one, dragging stuff around. Thanks!

system/core/pages.php
RoukePouw: Will do :)

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -69,7 +73,7 @@ WYMeditor.editor.prototype.image_upload = function() {
                     '<div class="row row-indent">' +
                         // We use a hidden value here so we can get a proper translation
                         '<input type="hidden" id="image_upload_uploading_label" value="{Uploading}" />' +
-                        '<input type="submit" class="submit" ' + 'value="{Upload}" />' +
+                        '<input id="upload-image" type="submit" class="submit" ' + 'value="{Upload}" />' +
```
matthijskooijman, line 76: I think this extra id is no longer needed.

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
matthijskooijman: Looking at the file contents, the original file uses tabs for indentation, but your inserted lines use 4 spaces. Might be good to fix in your editor.

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
matthijskooijman: Fixed.

system/core/editor.php
```diff
@@ -81,7 +81,8 @@ function loadEditor($html) {
 				Customizes the XHTML structure of WYMeditor.
 				"CONTAINERS" has been moved from "wym_area_right" to "wym_area_top":
 			*/
-			boxHtml: "<div class='wym_box'>\n"
+			boxHtml:
+                "<div class='wym_box'>\n"
```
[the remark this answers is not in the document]
matthijskooijman: Fixed.

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
matthijskooijman: Fixed.

matthijskooijman:
I fixed the issues I pointed out, squashed everything into one commit with a longer commit message (also adding a reference to the issue being fixed to allow auto-closing it) and rebased on top of master. I think this can be merged as-is now. Maybe @RoukePouw or @laurensmartina want to give this one last double-check?

RoukePouw:
> Maybe @RoukePouw or @laurensmartina want to give this one last double-check?

Double checked, looks good! Thanks for the final fixes!

matthijskooijman:
I fixed one more minor whitespace issue in the javascript, I'll merge this now.

Issue #312, opened 2020-05-12
PHP 7.4 crash?


tammoterhark:
I am not sure, don't know how to assess, but since my upgrade from Ubuntu 19.04 to 20.04 (and upgrade from PHP 7.2 to 7.4) my local installations of Hypha (both localhost.hypha.nl and localhost.destadsbron.nl) give me a completely white screen when starting. Other sites do work OK.

How to test/troubleshoot?

tammoterhark:
From the error log:

```
[Tue May 12 11:18:34.064934 2020] [php7:error] [pid 1316] [client 127.0.0.1:48504] PHP Fatal error:  Access level to DOMWrap\\Traits\\ManipulationTrait::inputAsNodeList() must be public (as in class DOMWrap\\Traits\\TraversalTrait) in /home/tammo/htdocs/hypha.nl/system/php-dom-wrapper/Element.php on line 75
[Tue May 12 11:18:34.161999 2020] [php7:error] [pid 1317] [client 127.0.0.1:48506] PHP Fatal error:  Access level to DOMWrap\\Traits\\ManipulationTrait::inputAsNodeList() must be public (as in class DOMWrap\\Traits\\TraversalTrait) in /home/tammo/htdocs/hypha.nl/system/php-dom-wrapper/Element.php on line 75
```

matthijskooijman:
@Tammo, I've pushed a fix to #313, could you see if it works for you with this change? Easiest is to just test the `fix-php7.4` branch.

tammoterhark:
I tested the patch and it works. Could this please be made live on De Stadsbron ASAP, since I am very handicapped at the moment because of this problem.

tammoterhark:
I managed to just change the code. So I can work now. But getting this live would be a help. No pressure.

matthijskooijman:
> Could this please be made live on De Stadsbron ASAP, since I am very handicapped at the moment because of this problem.

I'm not sure how this is relevant exactly? The Stadsbrton live server does not have PHP 7.4, so it works just fine. If you want to do some work locally, you can just checkout the fixed branch and work on top of that?

Regardless, I'll merge the PR (but might not push it to the Stadsbron live directly, since I think there is no real point there).

PR #313, opened 2020-05-12
Backport php-dom-wrapper changes for 7.4 compatibility

matthijskooijman:
This removes some abstract method definitions, which were defined public
here, but protected elsewhere. On PHP7.4, this produces:

    PHP Fatal error:  Access level to DOMWrap\\Traits\\ManipulationTrait::inputAsNodeList() must be public (as in class DOMWrap\\Traits\\TraversalTrait)

Since these abstract definitions are apparently not needed, just remove
them. This change is backported from upstream commit:

    https://github.com/scotteh/php-dom-wrapper/commit/8d332bf

Fixes: #312


matthijskooijman:
Tested by @tammoterhark, so merging.

PR #314, opened 2020-05-15
Issue/66 paste from word

RoukePouw:


system/core/editor.php
```diff
@@ -149,7 +149,16 @@ function loadEditor($html) {
 					this.isFocusSet = true;
 				}
 
-				// initiate the "image_upload" plugin
+                wym._iframe.contentDocument.addEventListener('paste', (event) => {
+                    const pasteContent = (event.clipboardData || window.clipboardData).getData('text');
+                    const textarea = document.createElement('TEXTAREA');
+                    textarea.value = pasteContent;
+                    wym.paste(textarea.value);
```
matthijskooijman, line 156: I actually suspect this textarea can be removed and you can just do `wym.paste(pasteContent)`. It looks like the `getData('text')` will already give the plaintext version of the clipboard contents, so going through the `textarea.value` does not do anything special (that's just a string variable essentially).

system/core/editor.php
```diff
@@ -149,7 +149,16 @@ function loadEditor($html) {
 					this.isFocusSet = true;
 				}
 
-				// initiate the "image_upload" plugin
+                wym._iframe.contentDocument.addEventListener('paste', (event) => {
```
matthijskooijman, line 152: There's again whitespace issues here (you're inserting spaces, while it should be tabs).

What editor are you using? We could maybe add a `.editorconfig` file to the project if your editor supports that (though I'm not sure if the indentation style is sufficiently consistent throughout the project...).

system/core/editor.php
RoukePouw: I suspect it as well, but couldn't test it so tried to replicate it as faithfully as possible for now.

matthijskooijman: I'm pretty sure that if it works for any other formatted text (e.g. from libreoffice, or a browser), it will work from Word too. The existing "Paste from Word" is actually not really Word-specific, it's just named that way because that's where the biggest junk comes from, I guess.

system/core/editor.php
laurensmartina: fixed

system/core/editor.php
laurensmartina: fixed

matthijskooijman:
This is superseded by #345, so I'm closing this one.

PR #316, opened 2020-05-15
Session improvements

matthijskooijman:
This PR improves the way sessions are handled. In particular:
 - It gracefully handles session start failures due to permission problems
 - It only creates a session for users that actually need it (i.e. not for regular visitors).
 - The session cookie is now better separated and secured
 - The CSRF token is moved from the session to a cookie (so not everyone will need a session anyway). This also means visitors that take a long time to type a comment are no longer met with a CSRF error.
 - The static `Hypha` class (that contained `HyphaFile`s) is merged into `RequestContext` (the former was a nice idea, but predated the latter and the latter is probably a better place).
 - A new `HyphaSession` class encapsulates all session access.
 - Some more `$_POST` accesses were converted to use `HyphaRequest` instead.
- Plus some more small things.

The first two points where what I originally set out to achieve, the other improvements and refactorings were either on the path those, were made possible by these refactorings, or did I just think of while working on this code.


matthijskooijman:
One of the changes this PR made was to *not* write the session on every request if it was not changed. However, I realized that this could actually cause the session to expire (session file deleted by the PHP session GC), because that checks for modified time rather than access time. This would mean the session lifetime was counted from login, rather than last access, which did not seem a good idea. So I removed that change and instead documented that the writing is intentional instead.

matthijskooijman:
I just rebased this on top of master, now #331 is merged (which removes a few commits from this PR).

system/core/base.php
```diff
@@ -310,7 +273,8 @@ function hypha_setTitle($string) {
 		returns hypha base html file contents
 	*/
 	function hypha_getHtml() {
-		return Hypha::$data->html->read();
+		global $O_O;
+		return $O_O->data->html->read();
```
laurensmartina, line 277: This should be `themeHtml` in stead of `html`; `$O_O->data->themeHtml->read();`

system/core/base.php
```diff
@@ -321,15 +285,17 @@ function hypha_getHtml() {
 		$contents - html file contents
 	*/
 	function hypha_setHtml($contents) {
-		Hypha::$data->html->writeWithLock($contents);
+		global $O_O;
+		$O_O->data->html->writeWithLock($contents);
```
laurensmartina, line 289: This should be `themeHtml` in stead of `html`; `$O_O->data->themeHtml->writeWithLock($contents);`

system/core/RequestContext.php
```diff
@@ -19,6 +19,12 @@ class RequestContext {
 		/** @var null|string */
 		private $csrfToken;
 
+		/**
+		 * Collection of various data files to be used in this
+		 * request.
+		 * @var StdClass */
```
laurensmartina, line 28: The `*/` should be on the next line when it comes to comment blocks.

system/core/base.php
```diff
@@ -321,15 +285,17 @@ function hypha_getHtml() {
...
 	}
 
 	/*
 		Function: hypha_getCss
 		returns hypha base css file contents
 	*/
 	function hypha_getCss() {
-		return Hypha::$data->css->read();
+		global $O_O;
+		return $O_O->data->css->read();
```
laurensmartina, line 298: This should be `themeCss` in stead of `css`; `$O_O->data->css->read();`

system/core/base.php
```diff
@@ -340,7 +306,8 @@ function hypha_getCss() {
 		$contents - css file contents
 	*/
 	function hypha_setCss($contents) {
-		Hypha::$data->css->writeWithLock($contents);
+		global $O_O;
+		$O_O->data->css->writeWithLock($contents);
```
laurensmartina, line 310: This should be `themeCss` in stead of `css`; `$O_O->data->css->writeWithLock($contents);`

system/core/HyphaSession.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+	/**
+	 * HyphaSession manages the PHP session for the current request.
+	 * It should always be used instead of direct access to
+	 * $_SESSION.
+	 */
+	class HyphaSession {
```
laurensmartina, line 8: To ensure the lock it would be better to apply the singleton pattern;


```
class HyphaSession {
    private static $instance;
    public static function getInstance() {
        if (!self::$instance instanceof HyphaSession) {
            self::$instance = new HyphaSession;
        }

        return self::$instance;
    }

    private function __construct() {
        // ....
    }
}

$hyphaSession = HyphaSession::getInstance();
```

system/core/base.php
matthijskooijman: Fixed.

system/core/base.php
matthijskooijman: Fixed.

system/core/base.php
matthijskooijman: Fixed.

system/core/base.php
matthijskooijman: Fixed.

system/core/HyphaSession.php
matthijskooijman: Done.

matthijskooijman:
Rebased on top of master, fixed a few things and added one more commit to prevent multiple instances of HyphaSession from being created. Ready to merge next.

PR #317, opened 2020-05-18
Fix cancel button on "forgot password" popup

### LINKED ISSUE #305: Cancel button on reset login prompt does not work
When I enter wrong credentials I get the message

"inloggen mislukt. Klik hier als je je wachtwoord bent vergeten: . heraanmelden"

when I click 'heraanmelden' I get a prompt to enter my email address or cancel, when I cancel nothing happens.

In the console I see this error:

ReferenceError: Can't find variable: showLogin

matthijskooijman:
This used to call `showLogin()` to show the login popup (since you
typically get to this popup from the login popup). However, the login
popup is shown using `login()`, rather than `showLogin()`.

This was already broken in the first git commit, so this probably never
worked, or got broken in a refactor way before that.

This fixes #305.

### DIFF
--- a/system/core/users.php
+++ b/system/core/users.php
@@ -148,7 +148,7 @@
 	function reregister() {
 		html = '<table class="section">';
 		html+= '<tr><th><?=__('name-or-email')?></th><td><input name="searchLogin" id="searchLogin" type="text" size="10" /></td></tr>';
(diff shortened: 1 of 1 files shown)


matthijskooijman:

PR #320, opened 2020-05-19
Improve dump function for NodeList and DomNode

matthijskooijman:
This makes some improvements to the `dump()` function that logs values to the javascript console from PHP.

In particular this improves the way that a `NodeList` and `DomNode` are dumped. For example, `dump($hyphaXml->find('user'));` would result in:

image

This shows the nodeList with an expandable elements list, and each DomNode in the list is shown as a native Javascript DomNode, which can be traversed interactively.

Without this PR, dumping a `NodeList` or `DomNode` would just show `Object { }` in the console.

### DIFF
--- a/system/core/events.php
+++ b/system/core/events.php
@@ -549,11 +549,9 @@
 		}
 		session_write_close();
 
-		if (count($hyphaDumpList)) {
-			foreach ($hyphaDumpList as $vars) {
-				foreach ($vars as &$value) $value = json_encode($value);
-				call_user_func_array([$html, 'writeScript'], ['console.log(' . implode(', ', $vars) . ');']);
-			}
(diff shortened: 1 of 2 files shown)


system/core/events.php
```diff
@@ -566,11 +566,9 @@ function addDumps($html) {
 		}
 		session_write_close();
 
-		if (count($hyphaDumpList)) {
-			foreach ($hyphaDumpList as $vars) {
-				foreach ($vars as &$value) $value = json_encode($value);
-				call_user_func_array([$html, 'writeScript'], ['console.log(' . implode(', ', $vars) . ');']);
-			}
+		foreach ($hyphaDumpList as $vars) {
```
laurensmartina, line 569: Way nicer!

matthijskooijman:

PR #329, opened 2020-06-11
Add support for macros

matthijskooijman:
This is still a WIP, just wanted to share my progress so far.


matthijskooijman:
I've now added a page list macro, which can be used to generate e.g. a homepage with recent articles, or a tag-based index page. Used e.g. like:

    <macro name="pagelist" tag="en/tag_label" pagetypes="peer_reviewed_article" languages="en" include-private="yes" sort="-date" limit="10" />


This probably still needs some more helper methods to make accessing macro options easier and more robust (though there is already some error checking in place).

matthijskooijman:
I think this is ready for merging now. I've used the pagelist macro to automate the homepage lists (both the audio/video as well as the list of normal articles) at http://test.destadsbron.nl/nl/home

Review of the code, but also the test site is welcome.

matthijskooijman:
I added some more options to the pagelist macro, it should now be suitable to be used on e.g. De Stadsbron to generate the homepage, tag pages and timeline/archive pages. I'll probably deploy this code already, while waiting for review and merging.

system/core/macros.php
```diff
@@ -0,0 +1,185 @@
...
+	 * defined in DateTime::createFromFormat). If the attribute is
+	 * not present, the given default is returned. If the value is
+	 * invalid, an UnexpectedValueException is raised, using the
+	 * given errorFormat to indicate the expected format to the
+	 * user.
+	 *
+	 * @param string $name The name of the attribute to get
+	 * @return DateTime | null
+	 */
+	protected function getDateTimeAttribute($name, $format, $errorFormat, $default = null) {
```
RoukePouw, line 83: I'm a bit surprised to see this function here. I wouldn't expect it to be part of the base class of a macro, seems a bit specific. But perhaps I'm missing something?

system/macros/pagelist.php
```diff
@@ -0,0 +1,219 @@
...
+		// Allow filtering by tag
+		$tagsattr = $this->macro_tag->getAttribute("tags");
+		$tags = self::lookupTags($tagsattr);
+
+		$excludeTagsattr = $this->macro_tag->getAttribute("exclude-tags");
+		$excludeTags = $this->lookupTags($excludeTagsattr);
+
+		// Include private pages only when logged in and enabled
+		// TODO: Better checking of option values (and maybe
+		// allow Yes/true/TrUe/etc.)
```
laurensmartina, line 128: I think there is a typo here. :-)

system/macros/pagelist.php
```diff
@@ -0,0 +1,219 @@
...
+
+		// Allow filtering by page type
+		$pageTypes = [];
+		$pageTypesAttr = $this->macro_tag->getAttribute("pagetypes");
+		if ($pageTypesAttr) {
+			$pageTypes = explode(',', $pageTypesAttr);
+			$validTypes = hypha_getDataTypes();
+			foreach ($pageTypes as $type) {
+				if (!array_key_exists($type, $validTypes))
+					throw new UnexpectedValueException("Invalid page type: $type. Valid options are: " . implode(', ', array_keys($validTypes)));
```
RoukePouw, line 139: Do we want to abort or fail with a message? If in the future one page type is removed/changed then the entire macro becomes broken.

matthijskooijman: This exception is caught further up, so it's just the macro that fails, not the entire page. That seems to me the safest and clearest way to fail, you could ignore the missing value and render something with a warning on top, but then it's really easy to miss the warning, so I think I'd prefer to just have the macro fail completely.

laurensmartina (approved):
Looks great!

system/core/macros.php
matthijskooijman: Well, I could imagine other macros that also need date/time parameters, so this is just a helper function to access such attributes/arguments. I also anticipate other helper functions for parsing other types of arguments, but I haven't had the need yet I guess.

Maybe it would make sense to move the actual parsing code to a more general place, and then here just call that?

matthijskooijman: I'm going to leave this for now and merge this. We can always move this code later if that turns out better (e.g. when we add a few more macros).

RoukePouw: Cool :)

system/macros/pagelist.php
matthijskooijman: Reading it again, I think that this is not a typo, but I meant to say that checking should be case insensitive :-)

matthijskooijman:
Rebased on top of master, and I added two more fixes:
 - An unrelated fix for empty index pages I came across
 - Letting the pagelist filter by the current language always, since this might produce fatal errors otherwise. This is temporary, until #310 is fixed.

With that, I'm merging this.

PR #330, opened 2020-06-23
Fix wikifying of root-relative links

matthijskooijman:
This makes sure that a link like `/en/pagename` is properly wikified too, rather than just `en/pagename` or `https://domain.tld/en/pagename`.

### DIFF
--- a/system/core/base.php
+++ b/system/core/base.php
@@ -704,10 +704,14 @@
 		make it a relative to $hyphaUrl.
 	*/
 	function hypha_make_relative($url) {
-		global $hyphaUrl;
-		$len = strlen($hyphaUrl);
-		if (substr($url, 0, $len) == $hyphaUrl)
-			$url = substr($url, $len);
+		global $O_O;
+
+		// Handle urls like https://mydomain.tld/path/en/pagename
+		$remove = $O_O->getRequest()->getRootUrl();
+		$len = strlen($remove);
+		if (substr($url, 0, $len) == $remove)
+			return substr($url, $len);
+
 		return $url;
 	}
 
--- a/system/core/base.php
+++ b/system/core/base.php
@@ -712,6 +712,12 @@
 		if (substr($url, 0, $len) == $remove)
 			return substr($url, $len);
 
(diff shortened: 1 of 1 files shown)


system/core/base.php
```diff
@@ -737,10 +737,19 @@ function hypha_make_absolute($url) {
...
 	*/
 	function hypha_make_relative($url) {
-		global $hyphaUrl;
-		$len = strlen($hyphaUrl);
-		if (substr($url, 0, $len) == $hyphaUrl)
-			$url = substr($url, $len);
+		global $O_O;
+
+		// Handle urls like https://mydomain.tld/path/en/pagename
+		$remove = $O_O->getRequest()->getRootUrl();
```
laurensmartina, line 743: Could this lead to an issue if the site can be accessed through multiple domains?
eg. http(s)://(www.)mydomain.tld/path/en/pagename
https://www.mydomain.tld/path/en/pagename
http://www.mydomain.tld/path/en/pagename
https://mydomain.tld/path/en/pagename
http://mydomain.tld/path/en/pagename

matthijskooijman: Yes, it could. However, this is already the case with the current code (that part is not really changed, just refactored to use `HyphaRequest::getRootUrl` rather than global `$hyphaUrl`, I just split this commit into two to emphasize this).

Also, this is only a problem when someone externally adds a URL to another version (another equivalent domain or http vs https) of the site, since all urls internally generated use the same url as is removed here. IOW, once the url is made relative once, it gets stored in the database relatively properly, and is potentially expanded to an absolute URL when e.g. editing the page, but can then always be made relative again when saving.

Also, we cannot really fix this properly now: The PHP code has no idea about the different domains that the site is available under. The only way AFAICS to fix this is to configure these domains manually, which makes this a lot more complex. So I would leave this problem unsolved for now, certainly within this PR>

matthijskooijman:
I rebased on top of master and split the last commit in two. The first commit has disappeared, since it was already merged in another PR. I would think this is ready to merge now.

PR #331, opened 2020-06-25
Store CSRF token in a cookie

matthijskooijman:
This is a subset of PR #316, with just the commits needed to move the CSRF token into a cookie. This was originally just a cleanup, but it has since been shown that having the CSRF token in the session can cause dataloss when visitors submit a (lengthy) comment and the session has expired since they first loaded the article page. In that case, the CSRF token will be invalid and the request rejected rather forcefully, losing the submitted comment.

Moving the CSRF token to a cookie ensures that the token survives until the browser closes, which should prevent this dataloss.

### DIFF
--- a/system/core/HyphaRequest.php
+++ b/system/core/HyphaRequest.php
@@ -134,6 +134,16 @@
 			return sprintf('%s://%s%s', $scheme, $host, $this->rootUrlPath);
 		}
 
+		/**
+		 * Returns the path component of the root url, i.e. the
+		 * part after the hostname. This always starts and ends
+		 * with a slash (which might be the same when running
+		 * in the root of the domain).
+		 */
+		public function getRootUrlPath() {
+			return $this->rootUrlPath;
+		}
+
 		public function getScheme() {
 			// Apache 2.4+ has REQUEST_SCHEME
 			if (array_key_exists('REQUEST_SCHEME', $_SERVER))
--- a/system/core/HyphaRequest.php
+++ b/system/core/HyphaRequest.php
@@ -148,7 +148,7 @@
 			// Apache 2.4+ has REQUEST_SCHEME
 			if (array_key_exists('REQUEST_SCHEME', $_SERVER))
 				return $_SERVER['REQUEST_SCHEME'];
-			if (array_key_exists('HTTPS', $_SERVER) && $_SERVER['HTTPS'] == 'on')
+			if ($this->isSecure())
 				return 'https';
 			return 'http';
 		}
@@ -159,6 +159,10 @@
 
 		public function getPort() {
 			return $_SERVER['SERVER_PORT'];
+		}
+
+		public function isSecure() {
+			return array_key_exists('HTTPS', $_SERVER) && $_SERVER['HTTPS'] === 'on';
 		}
 
 		/**
--- a/system/core/RequestContext.php
+++ b/system/core/RequestContext.php
@@ -16,6 +16,9 @@
 		/** @var array */
 		private $dictionary;
 
+		/** @var null|string */
+		private $csrfToken;
+
 		/**
 		 * @param HyphaRequest $hyphaRequest
 		 */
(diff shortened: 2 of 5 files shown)


system/core/events.php
```diff
@@ -256,55 +256,38 @@ function createGotoPageAndExecuteCommandButton($label, $langPageView, $command,
...
 			}
 		}
 	}
 
 	// execute posted commands
-	function executePostedCommand() {
-		if(isset($_POST['command'])) {
-			if (!isset($_POST['csrfToken']) || $_POST['csrfToken'] != getCsrfToken()) {
+	function executePostedCommand($O_O) {
+		static $processed = false;
```
RoukePouw, line 276: Would it make sense to add this global/static var to `$O_O`?
Feels weird to have such a state variable lying around :-)

system/core/HyphaRequest.php
```diff
@@ -151,6 +161,10 @@ public function getPort() {
 			return $_SERVER['SERVER_PORT'];
 		}
 
+		public function isSecure() {
+			return (array_key_exists('HTTPS', $_SERVER) && $_SERVER['HTTPS'] == 'on');
```
RoukePouw, line 165: r edundant parentheses (non blocking comment :-) )  maybe triple `===` ?

system/core/RequestContext.php
```diff
@@ -120,6 +124,62 @@ public function getDictionary() {
...
 		}
 
+		/*
+			Function: getCsrfToken
+
+			Returns the CsrfToken to be used for all
+			subsequent POST requests. If there is no token
+			yet, it will be generated automatically.
+		*/
+		public function getCsrfToken() {
```
RoukePouw, line 134: namingwise I would expect get to be just a getter (not meddling with state) the regeneration breaks that expectation. Maybe getOrGenerateCsrfToken / retrieveCsrfToken?

system/core/RequestContext.php
```diff
@@ -120,6 +124,62 @@ public function getDictionary() {
...
+
+			Returns whether the request is a POST request
+			that contains a valid CSRF token in the request
+			parameters. When this returns false, a POST
+			request should not be processed.
+		*/
+		public function validCsrfToken() {
+			$received = $this->getRequest()->getPostValue('csrfToken');
+			$expected = $this->getCsrfToken();
+			return ($received == $expected);
```
RoukePouw, line 180: redundant parentheses, maybe triple ===?

system/core/events.php
matthijskooijman: Hm, I'm not so sure. This sneakily hidden state is not ideal, exposing this state globally through `$O_O` does not seem correct either: This is really an implementation detail of `executePostedCommand`. I think it would be good to just keep it like this now, and fix this in a future refactor (the way post commands are handled is expected to change anyway, once all datatypes are converted to the "new style").

system/core/HyphaRequest.php
matthijskooijman: I like extra parenthesis to emphasize what is being returned, though thinking about it, I can't really put down in words why it would actually help :-) I guess it's a habit that's more relevant when assigning an expression to a variable, not really needed for return. Triple === is a good idea.

system/core/RequestContext.php
matthijskooijman: Good point, fixed.

system/core/RequestContext.php
matthijskooijman: Fixed.

system/core/HyphaRequest.php
matthijskooijman: Fixed.

matthijskooijman:
Thanks for the review. I've fixed most of your comments. I also noticed a missing `global $O_O;` that broke login (it got left behind splitting off commits into this new PR), that's also fixed now.

system/core/events.php
RoukePouw: Cool, just wanted to double check :-)

matthijskooijman:
I did a bit more testing, and will merge this now.

PR #337, opened 2020-08-19
Convert festival to new datatype style

matthijskooijman:
This improves the code style of the festival, as we previously did for other datatypes.

This PR starts with a few small refactorings that were easy to separate, and ends with one quite huge commit that applies a lot of changes to the festival module. These could maybe have been separated, but given the size of this refactor, this does not warrant the enormous amount of extra time that would have needed.

There are still some things that are imperfect (e.g. translation strings that get variables appended rather than interpolated), but these can better be fixed later than further enlarging this PR.

### DIFF
--- a/data/themes/default/hypha.css
+++ b/data/themes/default/hypha.css
@@ -862,6 +862,57 @@ textarea{
 .festivalTimetable .tableRowHeadingEven {
 	background-color:#bbb;
 }
+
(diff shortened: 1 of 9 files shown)


laurensmartina (approved):

matthijskooijman:
I squashed the fixup commits, should be ready to merge now.

PR #342, opened 2020-09-25
Featured image

laurensmartina:
adds featured image at peer_reviewed_article.


system/assets/hypha-preview-image-file-field.js
```diff
@@ -0,0 +1,64 @@
...
+
+					updateSrc(null);
+				})
+				.insertAfter($preview);
+		}
+
+		var toggle = function (hasSrc) {
+			if (hasSrc) {
+				$input.hide();
+				$('label[for="' + inputName + '"]').attr('for', inputName + '_disabled');
```
matthijskooijman, line 29: Why is this needed?

system/assets/hypha-preview-image-file-field.js
```diff
@@ -0,0 +1,64 @@
...
+
+			// update shadow file value
+			$('input[data-file-shadow-value-for="' + inputName + '"]').attr('value', src);
+
+			toggle(hasSrc);
+		}
+
+		var init = function () {
+			$input.on('change', function () {
+				updateSrc(window.URL.createObjectURL(this.files[0]));
```
matthijskooijman, line 56: Does this produce a meaningful URL, or a data url? If the latter, there's no real point in saving it into the shadow field in `updateSrc`, I think?

system/assets/hypha-preview-image-file-field.js
```diff
@@ -0,0 +1,64 @@
...
+			});
+
+			toggle($preview.attr('src'));
```
matthijskooijman, line 59: Why not merge `toggle` and `updateSrc` into one function?

I was also thinking hat if you use the value of the shadow field here, rather than the preview src, we could remove the preview handling on the PHP side, but that would prevent the preview from being resized (which might not be terrible, of course).

system/datatypes/peer_reviewed_article.php
```diff
@@ -1202,13 +1224,21 @@ protected function createEditForm(array $values = []) {
 			</div>
 EOF;
 
+		$src = array_key_exists(self::FIELD_NAME_FEATURED_IMAGE, $values) && is_string($values[self::FIELD_NAME_FEATURED_IMAGE]) && file_exists($values[self::FIELD_NAME_FEATURED_IMAGE]) ? $values[self::FIELD_NAME_FEATURED_IMAGE] : null;
+		$src = str_replace(HyphaImage::ROOT_PATH, '', $src);
```
matthijskooijman, line 1228: This would probably be more readable as an `if` statement.

system/datatypes/peer_reviewed_article.php
```diff
@@ -489,6 +498,12 @@ protected function editAction(HyphaRequest $request) {
 		$this->xml->find(self::FIELD_NAME_METHOD)->setHtml($form->dataFor(self::FIELD_NAME_METHOD));
 		$this->xml->find(self::FIELD_NAME_SOURCES)->setHtml($form->dataFor(self::FIELD_NAME_SOURCES));
 
+		$filename = $form->dataFor(self::FIELD_NAME_FEATURED_IMAGE);
+		if ($filename) {
+			$filename = HyphaImage::ROOT_PATH . $filename;
```
matthijskooijman, line 503: Why add the `ROOT_PATH` here? That only hurts portability (moving the hypha installation)? Leaving it out also prevents having to remove it again when inserting data into the form values later? Also, when using this filename, you'd normally pass it to the `HyphaFile` constructor, which also needs it *without* the `ROOT_PATH` prepended.

matthijskooijman (commented):
Looks good, very comfortable to review with all these nice and small commits :-)

I left some small remarks inline.

Furthermore:
 - I think that the festival module could also be updated to use the auto-generated shadowfield rather than defining its own shadow field (which is a bit weird because it mixes field names in the `handleImageUpload` call currently).
 - If the festival module is updated, I think that `handleImageUpload` could be generalized to a `handleFileUploads($_FILES)` that just processes *all* fields in `$form->file_fields`, rather than just one, so callers need less checking and less hardcoding of the form structure in the processing code.

system/languages/en.php
```diff
@@ -185,9 +185,9 @@
 		"invalid-money-amount" => "Not a valid amount of money",
 		"required-field-missing" => "Required field missing",
 		"failed-to-process-image" => "Failed to process image",
-		"image-type-must-be-one-of" => "Image type must be one of: ",
-		"error-uploading-file" => "Error uploading file: ",
-		"file-too-big-must-be-less-than" => "File is too big, it must be less than ",
+		"image-type-must-be-one-of" => "Image type must be one of: [[allowed-filetypes]]",
+		"error-uploading-file" => "Error uploading file: [[error]]",
+		"file-too-big-must-be-less-than" => "File is too big, it must be less than [[upload_max_filesize]]",
```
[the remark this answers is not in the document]
laurensmartina: good catch! fixed

system/languages/en.php
```diff
@@ -92,7 +92,7 @@
 		"page-name-conflict" => "page with this name already exists, sticking to old name",
 		"error-sending-message" => "failed to send email to the following address(es):",
 		"error-loading-html" => "setNodeHTML: error when loading HTML",
-		"too-big-file" => "Error: file must be less than ",
+		"too-big-file" => "Error: file must be less than [[upload-max-filesize]]",
```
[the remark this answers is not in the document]
laurensmartina: good catch! fixed

system/languages/en.php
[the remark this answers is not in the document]
laurensmartina: fixed

system/core/HyphaImage.php
```diff
@@ -77,10 +77,19 @@ function getFilename($width = 0, $height = 0) {
...
-		static function importUploadedImage($fileinfo, $max_size = 4194304 /* 4M */ ) {
-			if ($fileinfo['size'] > $max_size) return __('file-too-big-must-be-less-than', ['upload_max_filesize' => $max_size . 'bytes']);
-			if ($fileinfo['error'] == UPLOAD_ERR_INI_SIZE) return __('file-too-big-must-be-less-than', ['upload_max_filesize' => ini_get('upload_max_filesize')]);
-			if ($fileinfo['error']) return __('error-uploading-file', ['error' => $fileinfo["error"]]);
+		static function importUploadedImage($fileinfo) {
+			switch ($fileinfo['error']) {
+				case UPLOAD_ERR_INI_SIZE:
+					return __('file-too-big-must-be-less-than', ['upload_max_filesize' => ini_get('upload_max_filesize')]);
+				case UPLOAD_ERR_FORM_SIZE:
+					return 'The uploaded file exceeds the MAX_FILE_SIZE directive that was specified in the HTML form';
```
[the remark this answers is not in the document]
laurensmartina: fixed

system/core/HyphaImage.php
```diff
@@ -77,10 +77,19 @@ function getFilename($width = 0, $height = 0) {
 		 * If an error occurs a translated error message is
 		 * returned.
 		 */
-		static function importUploadedImage($fileinfo, $max_size = 4194304 /* 4M */ ) {
-			if ($fileinfo['size'] > $max_size) return __('file-too-big-must-be-less-than', ['upload_max_filesize' => $max_size . 'bytes']);
```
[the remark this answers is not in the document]
laurensmartina: Nope, placed it back

system/core/document.php
```diff
@@ -482,6 +504,13 @@ public function labelFor($name) {
 		}
 
 		public function dataFor($name, $default = null) {
+			if (!empty($this->file_fields[$name])) {
+				if (is_array($this->data) && !empty($this->data[$name])) {
```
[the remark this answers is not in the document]
laurensmartina: fixed

system/assets/hypha-preview-image-file-field.js
laurensmartina: I ran into the problem that the label was still working on the hidden input. I just fixed this issue by disabling the input. :-)

system/assets/hypha-preview-image-file-field.js
laurensmartina: True, i just need to put something there, so we can detect if the image remained the same, got replaced or removed.

system/assets/hypha-preview-image-file-field.js
laurensmartina: > I was also thinking hat if you use the value of the shadow field here, rather than the preview src, we could remove the preview handling on the PHP side
This would not work because the shadow field value does not contain the correct path.

system/assets/hypha-preview-image-file-field.js
matthijskooijman: What does "working on" mean for a label anyway? The label is still shown, and associated with the same input, which is hidden. Doesn't sound (semantically) incorrect? Does this confuse the browser somehow? You could argue that the label should now associate with the preview or the remove button, but I'm not so sure this is possible or useful?

system/assets/hypha-preview-image-file-field.js
matthijskooijman: But you do not need to detect "replaced" using the hidden field, I think, since that is detected when a new file upload has happened in the file input. Removed does need to be detected, but that can just empty out the field?

If we add a library-browser type thing later (where you can select existing images), you can just put the already known public URL of these images in the input field as well.

system/assets/hypha-preview-image-file-field.js
matthijskooijman: Yeah, I guess. Still, merging `toggle` and `updateSrc` seems reasonable?

system/assets/hypha-preview-image-file-field.js
```diff
@@ -0,0 +1,64 @@
+$(function() {
+	$('img[data-preview-for]').each(function () {
+		// get preview
+		var $preview = $(this);
+
+		// get input name
+		var inputName = $preview.data('preview-for');
+
+		// get input
+		var $input = $('input[name="' + inputName + '"][type="file"]');
```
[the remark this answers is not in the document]
laurensmartina: fixed

system/assets/hypha-preview-image-file-field.js
laurensmartina: fixed

system/assets/hypha-preview-image-file-field.js
laurensmartina: merged

system/datatypes/peer_reviewed_article.php
laurensmartina: fixed

system/datatypes/peer_reviewed_article.php
laurensmartina: If the `ROOT_PATH` is not saved, this whole part can be removed.

system/assets/hypha-preview-image-file-field.js
laurensmartina: It confusing for the user if a file upload dialog is triggered while that input field is hidden. I've added a comment. :-)

laurensmartina:
Processed all comments. @matthijskooijman please review the changes.

> I think that the festival module could also be updated to use the auto-generated shadowfield rather than defining its own shadow field (which is a bit weird because it mixes field names in the handleImageUpload call currently).

agreed

> If the festival module is updated, I think that handleImageUpload could be generalized to a handleFileUploads($_FILES) that just processes all fields in $form->file_fields, rather than just one, so callers need less checking and less hardcoding of the form structure in the processing code.

agreed


Let's fix the festival module after merging this PR.

system/assets/hypha-preview-image-file-field.js
matthijskooijman: Ah, I didn't realize that clicking the label triggers the input field, good point.

system/core/document.php
```diff
@@ -459,6 +471,16 @@ protected function updateFormField($field, $value) {
 			}
 		}
 
+		protected function addFileShadowValueField($field, $value) {
+			$name = $field->getAttribute('name');
+			$newField = $field->document()->createElement('input');
+			$newField->setAttribute('type', 'hidden');
+			$newField->setAttribute('value', $value);
+			$newField->setAttribute('name', 'file_shadow_' . $name);
+			$newField->setAttribute('data-file-shadow-value', '');
```
matthijskooijman, line 480: Why does this attribute exist?

matthijskooijman (approved):
I left one more question, and wondered: Should the featured image be shown along with the excerpt (for logged in users)?

system/core/document.php
laurensmartina: true, not needed. fixed

laurensmartina:
> I left one more question, and wondered: Should the featured image be shown along with the excerpt (for logged in users)?

Good idea! I'll implement it.

matthijskooijman:
@laurensmartina, I added one small simplification, see the added fixup commit. If that looks good to you, I'll squash and merge this tomorrow.

matthijskooijman:

PR #344, opened 2020-10-04
implement caption and meta-information for images #119

### LINKED ISSUE #119: implement caption and meta-information for images
title, attribution; store in title and longdesc attributes; display as figure with figcaption elements

autofill attribution from exif information

laurensmartina:
Fixes #119


system/core/pages.php
```diff
@@ -601,6 +601,19 @@ function wikify_link($node) {
 			$node->text('');
 	}
 
+	function figcaption($element) {
+		foreach ($element->findXPath(".//img[@longdesc]") as $img) {
+			figcaption_img($img);
+		}
+	}
+
+	function figcaption_img($img) {
```
matthijskooijman, line 610: I would rename these functions to more descriptive names, maybe `add_captions_to_all_images` and `add_caption_to_image`?

system/core/pages.php
```diff
@@ -601,6 +601,19 @@ function wikify_link($node) {
...
+		$text = $img->getAttribute('longdesc');
+		/** @var DOMWrap\Element $html */
+        $caption = new DOMWrap\Element('figcaption', $text);
+		$img->parentNode->insertBefore($caption, $img->nextSibling);
```
matthijskooijman, line 614: Doesn't this need a `<figure>` tag as well to group the caption with the image? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -45,12 +45,37 @@ WYMeditor.editor.prototype.image_upload = function() {
...
+		var wym = this,
+			options = wym._options,
+			imgAttrs,
+			selectedImage = wym.getSelectedImage();
+
+		imgAttrs = {
+			src: jQuery(options.srcSelector, wDialog.document).val(),
+			title: jQuery(options.titleSelector, wDialog.document).val(),
+			alt: jQuery(options.altSelector, wDialog.document).val(),
+			longdesc: jQuery('.wym_attribution', wDialog.document).val()
```
matthijskooijman, line 65: This seems weird: You enter an attribution, which is then saved as a "long description", and displayed as a caption, which seem three distinct things that I think we should not mix here? Maybe this is also something that @giplt could comment on?

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -45,12 +45,37 @@ WYMeditor.editor.prototype.image_upload = function() {
...
 					jQuery(".wym_src", doc).val(response.thumbUrl);
 					jQuery(".wym_alt", doc).val(response.original_filename);
+					jQuery(".wym_attribution", doc).val(response.attribution);
 				}
 				jQuery("form#image_upload_form .submit", doc).val(oldSubmitLabel);
 			}
 		})
 	};
 
+	d.submitHandler = function (wDialog) {
```
matthijskooijman, line 55: Why is this submitHandler completely new? I would think that adding attribution would only need to add one attribute to an existing handler. Is there maybe a hidden extra change here?

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
```diff
@@ -45,12 +45,37 @@ WYMeditor.editor.prototype.image_upload = function() {
 				} else {
 					jQuery(".wym_src", doc).val(response.thumbUrl);
 					jQuery(".wym_alt", doc).val(response.original_filename);
+					jQuery(".wym_attribution", doc).val(response.attribution);
```
matthijskooijman, line 48: IIUC this processes an "attribution" attribute returned from the image upload handler, but I see no changes in the upload handler?

matthijskooijman (commented):
I think I might be misunderstanding how all of this fits together, so I asked some questions inline. Maybe we should look at this together this afternoon.

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
laurensmartina: True! A next feature will be that the upload will return "attribution". This code is unaware of the state of the upload and already prepared.

system/core/pages.php
laurensmartina: fixed!

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
laurensmartina: I've implemented it like it was requested by @tammoterhark, I think he had a clear vision about this.

system/core/pages.php
laurensmartina: Ah, I didn't know! fixed

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
laurensmartina: No extra change, AFAIK the submitHandler needs to be overwritten in order to add a new attribute. Of course I'm open for suggestions.

system/wymeditor/lang/en.js
```diff
@@ -28,6 +28,8 @@ WYMeditor.STRINGS.en = {
     Title:            'Title',
     Relationship:     'Relationship',
     Alternative_Text: 'Alternative text',
+    Attribution:      'Attribution',
+    Attribution_placeholder: 'cc: author (date)',
```
matthijskooijman, line 32: I'm still unsure what "cc" means here. If it's intended to be "Creative Commons", then this might actualy be misleading, since just "CC" is not specific enough. I think we should just use "e.g." here, since it's intended as a suggestion? Or actually, maybe "license: author (date)" was the intention and is actually helpful, since it suggests to people that a license is actually needed. Let's try that and see what feedback we get.

matthijskooijman (changes requested):
I left a bunch of inline comments which I also fixed in two fixup commits that I'll push in a minute.

I did find some other problems that are still unfixed right now and might need us to rethink our approach. In particular:
 - The `<figure>` tag is a block-level tag, while `` is inline, so adding the `figure` creates invalid HTML, causing at least my Firefox to drop the `<figure>`, it seems.
 - Any layout CSS applied to the image might break. In particular, the default theme supports e.g. the `.left` class to make the image floating and have text wrap around it, but this breaks when wrapped inside `<figure>`.
 - Inside the wymeditor, the caption is not shown, which makes it a bit harder to preview changes.

Not sure if there is an easy fix for these, so maybe we should be storing `<figure>` tags after all?

matthijskooijman:
One additional thing I found is that you now explicitly exclude any HTML inside wymeditor, which I had expected to be not needed since this should live as text (rather than HTML) inside a textarea. However, I found that we actually changed this to HTML in b6f630ee8bb29a76691afe12b371ccd7c58a8987 to actually *allow* the global dewikify to process this. However, this results in actual (non-escaped) HTML to end up in the `<textarea>` tag, which is not technically correct HTML (though it seems browsers will allow it in practice).

Also note that this editor starts out as an `<editor>` tag and is later transformed into a `textarea` tag by [this bit of code](https://github.com/PlanBCode/hypha/blob/309d86e61890afaf49a5860be7be5dae5ee29f60/system/core/editor.php#L173-L187). This makes the current code in this PR fragile: If it would happen to run *before* rather than after this transformation, the `.wymeditor` class would not exist yet and it would add captions inside the wymeditor too.

I think it might be better to just revert to storing text rather than HTML, but manually run `dewikify` (and maybe other things needed) on the HTML before converting it to text and inserting it into the editor.

I also wonder if we really need this `editor` tag at all, maybe it would be simpler and more transparent to just use `<textarea class="wymeditor">` instead of `<editor>` in the forms directly? Does not really look any less readable to me either?

system/wymeditor/plugins/image_upload/jquery.wymeditor.image_upload.js
laurensmartina: updated to "attribution"

PR #349, opened 2020-10-13
mailinglist: Reset PHP timeout before every outgoing mail

matthijskooijman:


matthijskooijman:
Since this is a small change, and has already been deployed on the Stadsbron for a while now without problems, I'm going to merge this without review. I just rebased on top of master, will rebase next.

Issue #247, opened 2019-05-13
Force https


tammoterhark:
Using .htaccess:
https://serverguy.com/servers/redirect-http-to-https/

Both on live and test, so we test using the same parameters for this aspect...

matthijskooijman:
This should be optional (not all installations have https enabled). Since .htaccess is currently in git, this might not be the best place to handle this (and IIRC .htaccess does not support including a secondary ".htaccess.local" or similar). We could handle this in PHP instead with a checkbox in the settings.

It should probably only be possible to enable such a feature through a HTTPS request, to prevent locking out yourself by enabling the force HTTPS when HTTPS is not working yet. There is still the possibility of lockout when HTTPS breaks after enabling this. Perhaps the relevant settings page can be excluded from this, or people should just modify hypha.xml to recover (this requires documentation in both cases).

tammoterhark:
Logging in should also be possible in the situation that you describe.

tammoterhark:
And logging in should be really secure? Catch 22?

matthijskooijman:
Good point, so excluding the settings is not a good approach. Then it would be manually editing hypha.xml to recover (or, better, fix HTTPS again).

matthijskooijman:
One other approach would be to configure this using a special file. E.g. create a file called `system/force-https` or something like that, and check for that in .htaccess:

```
# Force HTTPS, if enabled by creating a specifically-named file:
RewriteCond %{HTTPS} off
RewriteCond system/force-https -f
RewriteRule ^(.*)$ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
```

This is untested, not sure if the relative path like this will always work. Also, not sure if this file should live inside system, or maybe in the root for extra visibility (or maybe be called FORCE-HTTPS,  or .FORCE-HTTPS to hide it, or...).

matthijskooijman:
One related thought is that in addition to forcing HTTPS, you might also want to force a specific domain when multiple aliases are available (e.g. always redirect example.com to www.example.com, or example.org to example.com). This is very similar to forcing HTTPS (including the same caveats about locking yourself out), so it would be good to implement this in the same way (possible using a single file, e.g. `system/force-url` that would contain `https://www.example.com` (but could maybe also contain `//www.example.com` to only force the domain or `https://` to only force the protocol?).

PR #351, opened 2020-10-20
Do not set cookies to https only

matthijskooijman:
This was only done for HTTPS-requests, but it turns out this breaks
mixed HTTP/HTTPS configurations, because browser do not store a separate
cookie for HTTP and HTTPS. So when you access the HTTPS version of a
site, the cookie is set as secure (https only) and subsequent HTTP
request will simply not get the cookie at all and cannot change the
cookie into non-secure either.

So this protection can only be enabled again when we know a site is
HTTPS-only, i.e. when #247 is implemented.

### DIFF
--- a/system/core/HyphaSession.php
+++ b/system/core/HyphaSession.php
@@ -29,7 +29,6 @@
 			// accessible to scripts).
 			session_name('hyphaSession');
 			ini_set('session.cookie_path', $request->getRootUrlPath());
-			ini_set('session.cookie_secure', $request->isSecure());
 			ini_set('session.cookie_httponly', true);
 			// This enables browser protections against CSRF
 			// attacks by omitting this cookie from all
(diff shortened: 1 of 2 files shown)


system/core/HyphaSession.php
```diff
@@ -29,7 +29,6 @@ public function __construct(HyphaRequest $request) {
 			// accessible to scripts).
 			session_name('hyphaSession');
 			ini_set('session.cookie_path', $request->getRootUrlPath());
-			ini_set('session.cookie_secure', $request->isSecure());
 			ini_set('session.cookie_httponly', true);
```
laurensmartina, line 32: Why not put the set secure within an if statement?

matthijskooijman: How do you mean exactly? Only set the property when `isSecure()` is true? That is what the old code effectively did, but that breaks mixed HTTP/HTTPS setups. Or am I misunderstanding you?

Writing this, I do realize that maybe we could include http/https in the session name so really support two different sessions for http and https *and* have secure cookies?

matthijskooijman:
I came up with a better approach, see #368.

PR #352, opened 2020-10-20
peer_reviewed_article: Show review checklist

### LINKED ISSUE #304: Automatic completeness of articles before starting review
Currently, excerpts in article are completely optional. It would be good if an excerpt is required to move the status from draft to ready for review.

I can imagine a few options:
 1. Show message *instead* of a "start review" button.
 2. Leave out the button and show a message "No excerpt yet, cannot start review until one is added" at the place where the excerpt
 3. Show an error message when clicking "start review" without an excerpt.

matthijskooijman:
In the DRAFT, REVIEW and APPROVED stages, this shows a checklist of things to do before progressing to the next stage, which helps to quickly see if anything is forgotten.

Some of these things are marked as required, meaning they prevent progress (i.e. starting review or publishing). Others are marked as recommended, meaning they are shown but not enforced. Whether something is required or recommended is hardcoded for now, but the code is writting to allow making this configurable later.

This additionally does some more or less related factorings and small cleanups.

The new functions have very verbose type definitions (intended to be consumed by Psalm), we will have to see if this is workable or simpler types should be used (which allow less checking).

Originally, I had intended for `updateToNewStatus` to also check the checklist and prevent state changes when the checklist was not completed, which would allow removing `canBeSetAsApproved` entirely (just try to set the status to `APPROVED` and if it fails, e.g. because the checklist says more approvals are needed, just ignore that failure). However, it turns out this could create a corner case where removing e.g. all tags after review start would prevent the status from progressing to `APPROVED`, so I ended up only checking the checklist in `statusChangeAction`. This results in code that is a bit more messy, but should be correct. @laurensmartina, any suggestions on this?

This fixes #304.


matthijskooijman:
Here's an overview of all the possible text that can be shown. Note that this includes situations that cannot occur with the current code or even do not make real sense (i.e. article content will probably always be required), but I've added them all to allow future expansion.

### Here's how an empty draft article would look in practice:
image
image

### Here's how a complete draft article would look
image
image

### And here's the various ways an article under review can look
image
image

image
image

### Here's the various ways a draft article could look with different settings
image
image

image
image

system/datatypes/peer_reviewed_article.php
```diff
@@ -91,6 +91,8 @@ class peer_reviewed_article extends HyphaDatatypePage {
 	const CMD_DISCUSSION_CLOSED = 'discussion_closed';
 	const CMD_COMMENT = 'comment';
 
+	const APPROVES_NEEDED = 3;
```
laurensmartina, line 94: Perfect! Next step: make it configurable. :-)

system/core/tags.php
```diff
@@ -47,6 +47,25 @@ static function findPagesWithTags(HyphaTag $tag, array $pageTypes=[], bool $incl
...
+		 * Return the tags assigned to the given page list node
+		 *
+		 * @return list<HyphaTag>
+		 */
+		static function tagsForPageListNode(HyphaDomElement $pageListNode) {
+			$idFilters = [];
+			foreach ($pageListNode->findXPath('./tag/@id') as $idAttr)
+				$idFilters[] = '@xml:id=' . xpath_encode($idAttr->value);
+			$tags = [];
+			if ($idFilters) {
```
laurensmartina, line 61: Here there should be an empty check.

system/core/language.php
```diff
@@ -1,10 +1,10 @@
...
-	/*
-		Function: __
+	/**
 		Returns a string from the selected dictionary.
 
-		Parameters:
-		$msgid - identifier used as a key to fetch the corresponding string from the dictionary.
+		@param string $msgid identifier used as a key to fetch the corresponding string from the dictionary.
+		@param null|array $args arguments to be interpolated into translated string
+		@return string
```
laurensmartina, line 7: The comment block is not correctly formatted

```
/**
 * <--
 */
```

system/datatypes/peer_reviewed_article.php
```diff
@@ -374,6 +387,176 @@ protected function defaultView(HyphaRequest $request) {
...
+	 *     status: self::CHECK_STATUS_*,
+	 *     label: string,
+	 *   }>}
+	 */
+	protected function runChecks() {
+		$result = [];
+		$overallStatus = self::CHECK_STATUS_PASSED;
+
+		foreach ($this->getChecks() as $name => $check) {
+			if (!in_array($this->getStatus(), $check['statuses']))
```
laurensmartina, line 483: The status gotten from `getStatus()` does not change while being in this method, so it would be better to assign in to a variable outside the foreach loop.

system/datatypes/peer_reviewed_article.php
```diff
@@ -374,6 +387,176 @@ protected function defaultView(HyphaRequest $request) {
...
+		}
+		$order = [self::CHECK_STATUS_PASSED, self::CHECK_STATUS_FAILED, self::CHECK_STATUS_RECOMMENDED];
+		uasort($result, function($a, $b) use ($order) {return array_search($a['status'], $order) - array_search($b['status'], $order); });
+
+		return [$overallStatus, $result];
+	}
+
+	/**
+	 * Renders a checklist div containing the given check result (as
+	 * returne by runChecks).
```
laurensmartina, line 511: Typo at `returne`

system/datatypes/peer_reviewed_article.php
```diff
@@ -719,7 +719,12 @@ protected function statusChangeAction(HyphaRequest $request, $newStatus) {
 			return ['redirect', $this->constructFullPath($this->pagename)];
 		}
 
-		$success = $this->updateToNewStatus($newStatus);
+		list($checkStatus, $_checks) = $this->runChecks();
+		if ($checkStatus == self::CHECK_STATUS_FAILED) {
```
laurensmartina, line 723: Strict comparison is recommended.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1793,22 +1793,11 @@ protected function getStatus() {
...
-		$blockingDiscussionsCount = $blockingDiscussions->count();
-		if ($blockingDiscussionsCount > 0) {
-			return false;
-		}
-
-		$approveCount = $this->xml->find(self::FIELD_NAME_APPROVE_CONTAINER)->children()->count();
-		if ($approveCount < self::APPROVES_NEEDED) {
-			return false;
-		}
+		list($_checkStatus, $checks) = $this->runChecks();
```
laurensmartina, line 1796: didn't know a better place to put this comment.. There is a typo in the commit message

laurensmartina (changes requested):

system/core/language.php
matthijskooijman: Fixed.

system/core/tags.php
matthijskooijman: Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Good call, fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Still not fully in my system ;-)

Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Found it, fixed :-)

system/datatypes/peer_reviewed_article.php
matthijskooijman: Yup :-)

matthijskooijman:
Fixed all comments, I think this should be ready to merge now (after squashing the fixup commits).

matthijskooijman:
I changed the checks slightly, so an article body with tags, but with an element like `img` or `iframe` is also acceptable. For the other fields (excerpt, sources and method), text is still required.

matthijskooijman:
I squashed the fixups and rebased on master, merging next.

PR #354, opened 2020-10-26
Allow moderation of public article comments

matthijskooijman:


matthijskooijman:
I further improved this PR, replacing the predefined reasons with a free-form input for reasons (which does remove the translations), allowing comments to be deleted (and undeleted) as well, and made the editing interface a bit more adaptive (i.e. change wordings based on whether a comment is already moderated or not).

There are a few commits that modify HTMLForm, adding methods related to select dropdowns, which are now no longer used. But since they're likely to be useful in the future, I've left these commits in the PR.

I did a force push for easier review of the complete changes.

system/core/document.php
```diff
@@ -514,6 +514,28 @@ public function findField($name, $tagName) {
...
+		 * by findField), using the given parameters.
+		 *
+		 * @param HyphaDomElement $field
+		 * @param string $value
+		 * @param string $content
+		 * @param array $attributes
+		 *
+		 * @return HyphaDomElement The created option element
+		 */
+		public function addOptionToSelect($field, $value, $content, $attributes = []) {
```
laurensmartina, line 528: use type hinting at `$field`.

laurensmartina: and also at `$attributes`

system/datatypes/peer_reviewed_article.php
```diff
@@ -1054,6 +1054,30 @@ protected function createApprovesDomElement() {
...
 
+	/**
+	 * Render a single comment.
+	 *
+	 * @param HyphaDomElement $comment
+	 * @param bool $review
+	 * @param bool $firstAndblocking
+	 * @return DomNodeList
+	 */
+	protected function createCommentDomElement($comment, $review, $firstAndBlocking) {
```
laurensmartina, line 1253: use typehinting

system/datatypes/peer_reviewed_article.php
```diff
@@ -1054,6 +1054,30 @@ protected function createApprovesDomElement() {
...
+		$createdAt = date('j-m-y, H:i', ltrim($comment->getAttribute(self::FIELD_NAME_CREATED_AT), 't'));
+		$committerName = $this->getCommentCommenter($comment);
+		$commentHtml = nl2br(htmlspecialchars($comment->getText()));
+		$commentHtml .= '<p>' . __('art-by') . ' <strong>' . htmlspecialchars($committerName) . '</strong> ' . __('art-at') . ' ' . htmlspecialchars($createdAt);
+		if (!$review && isUser()) {
+			$committerEmail = $comment->getAttribute(self::FIELD_NAME_DISCUSSION_COMMENTER_EMAIL);
+			$commentHtml .= ' <span> | ' . __('art-email') . ': <a href="mailto:' . htmlspecialchars($committerEmail) . '">' . htmlspecialchars($committerEmail) . '</a></span>';
+		}
+		if ($firstAndBlocking) {
+			$commentHtml .= ' | ' . ($resolved ? __('art-is-resolved') : __('art-is-blocking'));
```
laurensmartina, line 1263: `$resolved` is never passed

system/datatypes/peer_reviewed_article.php
```diff
@@ -778,6 +799,153 @@ protected function discussionClosedAction(HyphaRequest $request) {
...
 	}
 
+	/**
+	 * Constructs the WymHTMLForm for the comment form.
+	 *
+	 * @param HyphaDomElement $comment
+	 * @param array $values
+	 * @return HTMLForm
+	 */
+	protected function createCommentModerateForm($comment, $values = []) {
```
laurensmartina, line 812: use type hinting

system/datatypes/peer_reviewed_article.php
```diff
@@ -778,6 +799,153 @@ protected function discussionClosedAction(HyphaRequest $request) {
...
+	protected function commentModerateView(HyphaRequest $request) {
+		if (!$this->O_O->isUser()) {
+			notify('error', __('login-to-perform-action'));
+			return ['redirect', $this->constructFullPath($this->pagename)];
+		}
+
+		/** @var HyphaDomElement|null $comment */
+		$commentId = $request->getArg(1);
+		$comment = $this->xml->document()->getElementById($commentId);
+		if (!$comment instanceof HyphaDomElement || $comment->tagName != self::FIELD_NAME_DISCUSSION_COMMENT) {
```
laurensmartina, line 845: use strikt comparison `!==`

system/datatypes/peer_reviewed_article.php
```diff
@@ -1099,21 +1123,14 @@ protected function createDiscussionsDomElement($type) {
...
+
+				$li = $this->html->createElement('li')->appendTo($commentList);
+				$commentHtml = $this->createCommentDomElement($comment, $review, $firstComment && $blocking);
+				$li->append($commentHtml);
+
 				if ($firstComment) {
-					if ($blocking) {
-						$commentHtml .= ' | ' . ($resolved ? __('art-is-resolved') : __('art-is-blocking'));
-					}
+					$li->addClass("first-comment");
```
laurensmartina, line 1353: It would be more diverse if the class would be separated so that all comments get the "`comment`" class and the first comment would get "`first`" as well.

system/datatypes/peer_reviewed_article.php
```diff
@@ -778,6 +799,153 @@ protected function discussionClosedAction(HyphaRequest $request) {
...
+
+		$this->createCommentDomElement($comment, false, false)->appendTo($commentDiv);
+
+		$main->append($form);
+
+		$commands = $this->html->find('#pageEndCommands');
+		$commands->append($this->makeActionButton(__('cancel')));
+
+		$path = str_replace('{id}', $comment->getId(), self::PATH_COMMENTS_MODERATE);
+		$hide_title = $hidden ? __('art-moderate-change-reason') : __('art-moderate-hide-comment');
```
laurensmartina: the name of this variable is a bit confusion, it's not always hidden.

system/datatypes/peer_reviewed_article.php
```diff
@@ -778,6 +799,153 @@ protected function discussionClosedAction(HyphaRequest $request) {
...
+			notify('error', __('not-found')); // 404
+			return null;
+		}
+
+		$digestArgs = [
+			'commenter' => $comment->getAttribute(self::FIELD_NAME_DISCUSSION_COMMENTER_NAME),
+			'moderator' => $this->O_O->getUser()->getAttribute('fullname'),
+			'title' => $this->getTitle(),
+		];
+
```
laurensmartina, line 936: remove the second blank line

laurensmartina (changes requested):

system/core/document.php
matthijskooijman: Right, I did the comments, but then it is still useful to use PHP-readable hints as well, of course.

Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: w00ps, fixed :-)

system/datatypes/peer_reviewed_article.php
matthijskooijman: Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Maybe, or maybe using `:first-child` could remove the need for this class entirely. However, this just moves the existing class, so I'm going to consider this out of scope for the PR for now.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Do you mean that "hidden" is confusing in general, because you can still show the comment by expanding it? This is true, but I couldn't think of any better name ("hidden-by-default" is too complicated, I think). And it *is* hidden, it's just also
possible to reveal what's hidden :-)

Or do you mean that this particular button does not always hide the comment, it sometimes only changes the reason? That's not really what happens in practice, though, since this button always uses the HIDE command, so it always sets the moderate-status to hidden and sets the moderate-reason. So if the status is already hidden, this (in the view of the user) only changes the reason, but underwater it still does both.

Maybe this clarifies things? If not, let me know what you mean exactly, and maybe you also have a suggestion to improve?

system/datatypes/peer_reviewed_article.php
matthijskooijman: Fixed.

system/datatypes/peer_reviewed_article.php
matthijskooijman: Fixed.

system/datatypes/peer_reviewed_article.php
laurensmartina: agreed

system/datatypes/peer_reviewed_article.php
laurensmartina: I think I misread the `$hidden = $comment->getAttribute(self::FIELD_NAME_DISCUSSION_MODERATION_STATUS) === self::MODERATION_STATUS_HIDDEN;` for `$hidden = $comment->getAttribute(self::FIELD_NAME_DISCUSSION_MODERATION_STATUS);`.
My bad.

laurensmartina (approved):

matthijskooijman:
Thanks! I squashed the fixups and will merge now.

Issue #332, opened 2020-08-03
Count down for mailinglist


giplt:
Currently, a mailing list page needs to stay opened when sending out a newsletter. As subscriptions grow this takes more time. A count down counter both draws attention to this fact and helps a user in knowing when it's safe to leave the page.

PR #357, opened 2020-12-04
Display progress for mailinglist send

RoukePouw:
#332

Extend makeAction function with javascript callback
Set intermediate progress updates in xml
Lock and unload browser page with javascript.

### DIFF
--- a/system/core/various.php
+++ b/system/core/various.php
@@ -97,7 +97,7 @@
 	*/
 	function makeButton($label, $action, $id = '', $class = '') {
 		$class = 'button' . ($class ? ' ' . $class : '');
-		return '<input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'.$action.'" />';
+		return '<input type="button" class="'.htmlspecialchars($class).'" '.($id?' id="'.htmlspecialchars($id).'"':'').'value="'.htmlspecialchars($label).'" onclick="'.htmlspecialchars($action).'" />';
 	}
 
 	/*
--- a/system/datatypes/mailinglist.php
+++ b/system/datatypes/mailinglist.php
@@ -31,6 +31,8 @@ class mailinglist extends HyphaDatatypePage {
 	const FIELD_NAME_MESSAGE = 'message';
 	const FIELD_NAME_ADDRESS = 'address';
 	const FIELD_NAME_STATUS = 'status';
+	const FIELD_NAME_TOTAL = 'total';
+	const FIELD_NAME_PROGRESS = 'progress';
 	const FIELD_NAME_CONFIRM_CODE = 'confirm-code';
 	const FIELD_NAME_UNSUBSCRIBE_CODE = 'unsubscribe-code';
 	const FIELD_NAME_REMINDED = 'reminded';
@@ -60,6 +62,7 @@ class mailinglist extends HyphaDatatypePage {
 	const CMD_SAVE = 'save';
 	const CMD_SEND = 'send';
 	const CMD_SEND_TEST = 'test_send';
+	const CMD_PROGRESS = 'progress';
 
 	const MAILING_STATUS_DRAFT = 'draft';
 	const MAILING_STATUS_SENDING = 'sending';
@@ -69,6 +72,7 @@ class mailinglist extends HyphaDatatypePage {
 	const ADDRESS_STATUS_CONFIRMED = 'confirmed';
 	const ADDRESS_STATUS_UNSUBSCRIBED = 'unsubscribed';
 
+	const PROGRESS_UPDATE_INTERVAL_MS = 1000;
 	const TIME_LIMIT_PER_MAIL = 30;
 
 	public function __construct(HyphaDomElement $pageListNode, RequestContext $O_O) {
@@ -116,6 +120,7 @@ class mailinglist extends HyphaDatatypePage {
 			case [self::PATH_MAILS,                null]:                return $this->mailingView($request);
 			case [self::PATH_MAILS,                self::CMD_SEND]:      return $this->mailingSendAction($request);
 			case [self::PATH_MAILS,                self::CMD_SEND_TEST]: return $this->mailingSendTestAction($request);
+			case [self::PATH_MAILS,                self::CMD_PROGRESS]:  return $this->mailingProgressAction($request);
 			case [self::PATH_MAILS_EDIT,           null]:                return $this->mailingEditView($request);
 			case [self::PATH_MAILS_EDIT,           self::CMD_SAVE]:      return $this->mailingEditAction($request);
 		}
@@ -646,6 +651,11 @@ class mailinglist extends HyphaDatatypePage {
 		return null;
 	}
 
+	protected function getNumberOfAddresses(){
+		$xpath = './/' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_STATUS . '="' . self::ADDRESS_STATUS_CONFIRMED . '"]';
+		$addresses = $this->findAddresses($xpath);
+		return count($addresses);
+	}
 	/**
 	 * @param HyphaRequest $request
 	 *
@@ -697,14 +707,58 @@ class mailinglist extends HyphaDatatypePage {
 		if (self::MAILING_STATUS_DRAFT == $status) {
 			$editPath = $this->substituteSpecial(self::PATH_MAILS_EDIT_ID, ['id' => $mailing->getId()]);
 			$commands->append($this->makeActionButton(__('edit'), $editPath));
-			$xpath = './/' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_STATUS . '="' . self::ADDRESS_STATUS_CONFIRMED . '"]';
-			$addresses = $this->findAddresses($xpath);
-			$num = count($addresses);
-
+			$num = $this->getNumberOfAddresses();
 			$linkToMailing = $this->path(self::PATH_MAILS_VIEW_ID, ['id' => $mailingId]);
-			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . '))' . makeAction($linkToMailing, self::CMD_SEND, '');
-			$action = str_replace(['"'], ['\''], $action);
-			$commands->append(makeButton(__('send'), $action));
+
+			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . ')){
+				const divPopUp = document.createElement("div"); // create a popup
+				divPopUp.className = "messagebox modal";
+				const divProgress = document.createElement("div");
+				divProgress.className = "progress";
+
+				const spanProgressText = document.createElement("span");
+				spanProgressText.innerHTML= "'.__('ml-send-in-progress').'";
+				spanProgressText.className = "progress-text";
+				const spanProgressIndicator = document.createElement("span");
+				spanProgressIndicator.className = "progress-indicator";
+				spanProgressIndicator.innerHTML = "0/'.$num.'";
+				const divWarning = document.createElement("div");
+				divWarning.className = "warning";
+				divWarning.innerHTML = "'.__('ml-do-not-close').'";
+				divProgress.appendChild(spanProgressText);
+				divProgress.appendChild(spanProgressIndicator);
+				divPopUp.appendChild(divProgress);
+				divPopUp.appendChild(divWarning);
+
+				const divGray = document.createElement("div"); // create a gray layer
+				divGray.className="gray-modal-background";
+				document.body.appendChild(divGray);
+				document.body.appendChild(divPopUp);
+				let checkProgressInterval;
+				const onCompleted = function(){
+					if(checkProgressInterval) clearInterval(checkProgressInterval);
+					checkProgressInterval = null;
+					if(divPopUp && divPopUp.parentNode === document.body) document.body.removeChild(divPopUp);
+					if(divGray && divGray.parentNode === document.body) document.body.removeChild(divGray);
+					window.onbeforeunload = undefined;
+				};
+				const updateProgress = function(result){
+					const {'.self::FIELD_NAME_STATUS.','.self::FIELD_NAME_PROGRESS.','.self::FIELD_NAME_TOTAL.'} = JSON.parse(result);
+					if('.self::FIELD_NAME_PROGRESS.'>='.self::FIELD_NAME_TOTAL.' || '.self::FIELD_NAME_STATUS.' === "'.self::MAILING_STATUS_SENT.'") onCompleted();
+					else spanProgressIndicator.innerHTML = '.self::FIELD_NAME_PROGRESS.'+"/"+'.self::FIELD_NAME_TOTAL.';
+				};
+				const checkProgress = function(){'. // retrieve the progress of the send action
+					makeAction($linkToMailing, self::CMD_PROGRESS, '', null, 'updateProgress')
+				.'};
+				checkProgressInterval = setInterval(checkProgress,'.self::PROGRESS_UPDATE_INTERVAL_MS.');
+				window.onbeforeunload = function(){return \''.__('ml-sure-to-close').'\';};' // lock the browser from exiting.
+				 . makeAction($linkToMailing, self::CMD_SEND, '', null, 'onCompleted').';
+			 }';
+
+			if ($num > 0){ // hide send button if no addresses available
+				$action = str_replace(['"'], ['\''], $action);
+				$commands->append(makeButton(__('send'), $action));
+			}
 			$action = 'hypha('.json_encode($linkToMailing).', '.json_encode(self::CMD_SEND_TEST).', prompt(' . json_encode(__('email')) . '), $(this).closest(\'form\'));';
 			$action = str_replace(['"'], ['\''], $action);
 			$commands->append(makeButton(__('ml-test-send'), $action));
@@ -1010,6 +1064,8 @@ EOF;
 		$this->xml->lockAndReload();
 		$mailing = $this->xml->getElementById($mailingId);
 		$mailing->setAttribute('status', self::MAILING_STATUS_SENDING);
+		$count = 0;
+		$mailing->setAttribute(self::FIELD_NAME_PROGRESS, $count);
 		/** @var HyphaDomElement $mailings */
 		$mailings = $this->getDoc()->getOrCreate('mailings');
 		$mailings->append($mailing);
@@ -1018,6 +1074,11 @@ EOF;
 		// iterate over the confirmed addresses and send the mailing
 		$xpath = './/' . self::FIELD_NAME_ADDRESS . '[@' . self::FIELD_NAME_STATUS . '="' . self::ADDRESS_STATUS_CONFIRMED . '"]';
 		$receivers = $this->findAddresses($xpath);
+		if (count($receivers) === 0) {
+			notify('error', __('ml-mail-has-invalid-status'));
+			return null;
+		}
+
 		$linkToMailing = $this->path(self::PATH_MAILS_VIEW_ID, ['id' => $mailingId]);
 		/** @var HyphaDomElement $receiver */
 		foreach ($receivers as $receiver) {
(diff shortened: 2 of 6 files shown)


system/datatypes/mailinglist.php
```diff
@@ -702,7 +702,13 @@ protected function mailingView(HyphaRequest $request) {
 			$num = count($addresses);
 
 			$linkToMailing = $this->path(self::PATH_MAILS_VIEW_ID, ['id' => $mailingId]);
-			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . '))' . makeAction($linkToMailing, self::CMD_SEND, '');
+
+			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . ')){
+				const checkProgress = ()=>{}; // TODO do a call to get the progress and display this in a bar
+				setInterval(checkProgress,1000); //TODO make into a constant
+				window.onbeforeunload = ()=>\'Are you sure you want to leave?\';' // lock the browser from exiting. TODO translation?
```
matthijskooijman, line 709: If this is shown to the user, it should probably be more explicit about what happens when you leave?

matthijskooijman (commented):
I had a quick look at this code, looks like it's heading in a good direction. Before the final merge, I think some things (at least the callback argument to `makeAction`) should be moved into separate commits, of course.

Also, this progress tracking could maybe also result in resumable mailinglist sending, I think (because subscribers are never deleted, only changes status when they unsubscribe IIRC, so the `progress` attribute can be used to pick up where an interrupted sending left off). Probably not in scope for this PR, but good to keep in mind :-)

matthijskooijman:
Also, be sure to put "Fixes: #332" in the commit message that actually implements the core of this PR, so that issue can be autoclosed.

system/datatypes/mailinglist.php
```diff
@@ -697,14 +704,60 @@ protected function mailingView(HyphaRequest $request) {
...
+			  const checkProgress = function(){'. // retrieve the progress of the send action
+					makeAction($linkToMailing, self::CMD_PROGRESS, '', null, 'updateProgress')
+				.'};
+				checkProgressInterval = setInterval(checkProgress,1000); //TODO make into a constant
+				window.onbeforeunload = function(){return \'Are you sure you want to leave?\';};' // lock the browser from exiting. TODO translation?
+				 . makeAction($linkToMailing, self::CMD_SEND, '', null, 'onCompleted').';
+			 }';
+
+			if ($num > 0){ // hide send button if no addresses available
+				$action = str_replace(['"'], ['\''], $action);
```
matthijskooijman, line 757: Why is this replace needed? If this is because `"` breaks the `onclick="..."` quoting, then maybe this should be fixed by using `htmlspecialchars` in `makeAction` instead.

matthijskooijman: Seems this is from existing code, but still interesting to look into this.

system/datatypes/mailinglist.php
```diff
@@ -1018,18 +1073,31 @@ protected function mailingSendAction(HyphaRequest $request) {
...
 			$linkToUnsubscribe = $this->path(self::PATH_UNSUBSCRIBE_CODE, ['address' => $email, 'code' => $code]);
 			$message = '<p><a href="' . $linkToMailing . '">' . __('ml-if-unreadable-use-link') . '</a></p>';
 			$message .= $mailing->getHtml();
 			$message .= '<p><a href="' . $linkToUnsubscribe . '">' . __('ml-unsubscribe') . '</a></p>';
-			$this->sendMail($mailing->getAttribute('subject'), $message, [$email], $senderData['email'], $senderData['name']);
+			//TODO disabled for testing purposes $this->sendMail($mailing->getAttribute('subject'), $message, [$email], $senderData['email'], $senderData['name']);
+
+			$this->xml->lockAndReload();
+			$mailing = $this->xml->getElementById($mailingId);
+			$mailing->setAttribute('progress', ++$count);
```
matthijskooijman, line 1097: Should the progress attribute be removed when the send is complete?

system/datatypes/mailinglist.php
```diff
@@ -1083,6 +1151,25 @@ protected function mailingSendTestAction(HyphaRequest $request) {
...
+
+		$mailingId = $request->getArg(1);
+
+		// check if given id was correct
+		/** @var HyphaDomElement $mailing */
+		$mailing = $this->xml->getElementById($mailingId);
+		if (!$mailing instanceof HyphaDomElement) {
+			return '404';
+		}
+		$progress = $mailing->getAttribute('progress').'/'.$this->getNumberOfAddresses();
```
matthijskooijman, line 1167: Would it be better to return JSON instead of a 0/0 format that needs to be split client-side? JSON is easier and more robust to parse?

system/datatypes/mailinglist.php
```diff
@@ -1083,6 +1151,25 @@ protected function mailingSendTestAction(HyphaRequest $request) {
 		return 'reload';
 	}
 
+	protected function mailingProgressAction(HyphaRequest $request) {
+		if (!$this->hasSender()) {
```
matthijskooijman, line 1154: Is this check really relevant here? Seems like a copy-paste that is not really needed here?

matthijskooijman (commented):
Looks good. Left some comments inline (I think you already fixed translations, so I'll resolve those comments directly if they're fixed).

Overall:
 - This should be split over multiple commits.
 - There is a *lot* of inline JS. I wonder if that would be better in a script tag, which can be done using the HTMLDocument `writeScript()` method.

system/languages/en.php
```diff
@@ -313,6 +313,11 @@
 		"ml-received-by" => "received by",
 		"ml-sure-to-send" => "Are you sure you want to send off this mailing to [[count]] recipient(s)?",
 
+		// mailing list - progress feedback ui
+		"ml-sure-to-close" => "Are you sure you want to leave and stop sending the mailing?",
+		"ml-do-not-close" => "Attention: Do not close this window untill all mails are sent.",
+		"ml-send-in-progress" => "The mail is being send...",
```
matthijskooijman, line 319:
```suggestion
		"ml-send-in-progress" => "The mail is being sent...",
```

system/datatypes/mailinglist.php
```diff
@@ -697,14 +704,60 @@ protected function mailingView(HyphaRequest $request) {
...
-			$num = count($addresses);
-
+			$num = $this->getNumberOfAddresses();
 			$linkToMailing = $this->path(self::PATH_MAILS_VIEW_ID, ['id' => $mailingId]);
-			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . '))' . makeAction($linkToMailing, self::CMD_SEND, '');
-			$action = str_replace(['"'], ['\''], $action);
-			$commands->append(makeButton(__('send'), $action));
+
+			$action = 'if(confirm(' . json_encode(__('ml-sure-to-send', ['count' => $num])) . ')){
+				const divPopUp = document.createElement("DIV"); // create a popup
```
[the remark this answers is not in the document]
RoukePouw: Fixed :)

system/datatypes/mailinglist.php
```diff
@@ -697,14 +704,60 @@ protected function mailingView(HyphaRequest $request) {
...
+				spanProgressIndicator.innerHTML = "0/'.$num.'";
+				const divWarning = document.createElement("div");
+				divWarning.className = "warning";
+				divWarning.innerHTML = "Let op: Sluit dit venster niet tot de hele mailing is verstuurd";
+				divProgress.appendChild(spanProgressText);
+				divProgress.appendChild(spanProgressIndicator);
+				divPopUp.appendChild(divProgress);
+				divPopUp.appendChild(divWarning);
+
+				const divGray = document.createElement(\'DIV\'); // create a gray layer
```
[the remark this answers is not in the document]
RoukePouw: Fixed :)

system/datatypes/mailinglist.php
```diff
@@ -697,14 +704,60 @@ protected function mailingView(HyphaRequest $request) {
...
+					window.onbeforeunload = undefined;
+				};
+				const updateProgress = function(progress){
+					const [done,total] = progress.split("/").map(Number);
+					if(done>=total) onCompleted();
+					else{
+						spanProgressIndicator.innerHTML = progress;
+					}
+				};
+			  const checkProgress = function(){'. // retrieve the progress of the send action
```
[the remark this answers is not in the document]
RoukePouw: Fixed :)

system/datatypes/mailinglist.php
RoukePouw: Then the progress checker will find no final progress '10/10' and cannot determine that it has finished (using the way it has been build now).

system/datatypes/mailinglist.php
RoukePouw: We can do that. Better/easier is a bit subjective  :-) 

JSON is used because is it both human readable and computer parsable. If find `'5/10'` more readable than `{"current":5,"total":10}`. 
Parsing is `const {current,total} = JSON.parse(string);` vs `const [current,total] = string.split('/');` not much difference in complexity there.

And as to robustness `(x,y) => x+'\'+y` is a fully bijective function as for x,y without slashes so parsing is as robust as it gets (and probably even faster than JSON parse)

But I'm more than fine to switch to JSON :-)

system/datatypes/mailinglist.php
RoukePouw: Yeah, I just kept it there. I can look into it if that's helpful?

system/languages/en.php
RoukePouw: Fixed :)

system/datatypes/mailinglist.php
RoukePouw: A copy paste  indeed. I kept it as it seemed to be used in other functions. It made sense to me that a mailing without a sender would fail to return progress. 
Should I remove it?

RoukePouw:
> There is a lot of inline JS. I wonder if that would be better in a script tag, which can be done using the HTMLDocument writeScript() method.

Now sure how that would work, can you explain?

In other cases I would define a file with the functions that would get packed together into a minified library or something. But with hypha it seems to be the convention to group code together in a single file.

RoukePouw:
> This should be split over multiple commits.

How would you like this done?

system/datatypes/mailinglist.php
matthijskooijman: Agreed, but JSON is also more extendable (and if we add other AJAX APIs, can be used more consistently), so I'd still favor JSON here.

system/datatypes/mailinglist.php
matthijskooijman: But the progress endpoint could just look at the status for that? Even more, if the endpoint is switched to JSON, it could maybe even return a "status" (or "completed") attributed as well?

system/datatypes/mailinglist.php
matthijskooijman: I suspect there is some fragility in the current system because of this, so if you want to have look, please do.

system/datatypes/mailinglist.php
RoukePouw: Agreed :)

system/datatypes/mailinglist.php
matthijskooijman: I would remove it for simplicity.

system/datatypes/mailinglist.php
RoukePouw: Yes, will do that :)

matthijskooijman:
> How would you like this done?

Do you mean "Through what process"? Or "split into which commits"?

RoukePouw:
> Do you mean "Through what process"? Or "split into which commits"?

First in which commits you would like it split. (and why :-) so I can understand the thoughts behind it ) Secondly, yes, if you have pointers on how to do this, otherwise I'll do some DuckDuckGo-ing myself ;-)

system/datatypes/mailinglist.php
RoukePouw: Done :)

system/datatypes/mailinglist.php
RoukePouw: And done

system/datatypes/mailinglist.php
RoukePouw: And done

system/datatypes/mailinglist.php
RoukePouw:
```
system/core/various.php:98

function makeButton($label, $action, $id = '', $class = '') {
		$class = 'button' . ($class ? ' ' . $class : '');
		return '<input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'.$action.'" />';
	}
```

I would suggest doing a fix here instead of in mailinglist.php. We could do something along these lines:

```
function makeButton($label, $action, $id = '', $class = '') {
		$class = 'button' . ($class ? ' ' . $class : '');
		$action = makeSave($action);
		return '<input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'.$action.'" />';
	}
```

but I wouldn't recommend using `htmlspecialchars`. Because that will make html entities out of javascript code, right?
Simply replacing `"` with `\'` could also lead to unpredictable behavior, depending on your javascript usage of quotes. Same with replacing `"` with `\"`. There's always an edge case that could be triggered.

Best way I can think of might be:

```
function makeButton($label, $action, $id = '', $class = '') {
		$class = 'button' . ($class ? ' ' . $class : '');
		$functionName = '?';
		return '<script>function '.$functionName.'(){'.$action.'}</script><input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'$functionName'();" />';
	}
```

But then we would need a proper way of determining a good unique name for `$functionName`;

matthijskooijman:
I would make commits to:
 - Remove the extra request headers
 - Support callback for `makeAction`
 - Fix for `str_replace(['"'], ['\''], $action);` if you find it
 - Optionally the CSS for the messagebox, because it is fairly general (but probably better to just include this in the next commit)
 - Everything else (i.e. implementing the progress counter)

As for the how: In this case I think none of the current commits are usable as is, so it is probably best to do a (soft) reset back to the original commit (just before your first commit), to effectively remove all commits, but leave the changes in your working copy, so you can use e.g. `git add -p` to create new commits from that. If you would want to preserve existing commits but split or rearrange some, then using an interactive rebase would be the way to go.

system/datatypes/mailinglist.php
matthijskooijman: > but I wouldn't recommend using htmlspecialchars. Because that will make html entities out of javascript code, right?

It would, but I don't think that's a problem. Since HTML attributes  (such as onclick) should be entity-decoded before they are used (this should be handled transparently by the DOM parser), so the javascript interpreter should see exactly the input the `htmlspecialchars`. One notable exception is CDATA (inside CDATA blocks or some tags), which is used literal, but that doesn't apply here AFAICS.

> Best way I can think of might be:

Clever approach, but it also seems a bit fragile. And indeed, it needs unique function names.

RoukePouw: htmlspecialchars it is :-)

RoukePouw:
Commits have been split :)

system/datatypes/mailinglist.php
RoukePouw: Though this does leave the edge case that, for example, `console.log('&gt;')` will not perform as expected.

matthijskooijman: Why not? Did you test this? I think this should be encoded as `console.log(&amp;gt;')` in the HTML, which should be decoded into the proper code again before running the javascript?

RoukePouw: My bad, apparently I tested it to hastily and messed up enabling the `htmlspecialchars` :-S

system/datatypes/mailinglist.php
```diff
@@ -1083,6 +1149,24 @@ protected function mailingSendTestAction(HyphaRequest $request) {
...
+		/** @var HyphaDomElement $mailing */
+		$mailing = $this->xml->getElementById($mailingId);
+		if (!$mailing instanceof HyphaDomElement) {
+			return '404';
+		}
+		$status = $mailing->getAttribute(self::FIELD_NAME_STATUS);
+		$progress = $mailing->getAttribute(self::FIELD_NAME_PROGRESS);
+		$total = $this->getNumberOfAddresses();
+		$result = [ self::FIELD_NAME_PROGRESS => $progress, self::FIELD_NAME_TOTAL => $total, self::FIELD_NAME_STATUS => $status];
+		echo json_encode($result);
```
matthijskooijman, line 1166: This should probably also set a Content-type header, to allow browsers and client frameworks to know that his is JSON. Not strictly required now, but probably helpful for debugging nonetheless.

system/datatypes/mailinglist.php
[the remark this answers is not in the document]
matthijskooijman: Nope, seems to still be there. The file is indented with tabs, and this line seems use tabs followed by one level of 2 spaces (which doesn't show up when you display tabs as 2 spaces, but is inconsistent and shows up with other tab stop settings).

system/datatypes/mailinglist.php
```diff
@@ -1083,6 +1149,24 @@ protected function mailingSendTestAction(HyphaRequest $request) {
 		return 'reload';
 	}
 
+	protected function mailingProgressAction(HyphaRequest $request) {
+
```
matthijskooijman, line 1153: Extra empty line.

system/datatypes/mailinglist.php
```diff
@@ -1010,6 +1062,8 @@ protected function mailingSendAction(HyphaRequest $request) {
 		$this->xml->lockAndReload();
 		$mailing = $this->xml->getElementById($mailingId);
 		$mailing->setAttribute('status', self::MAILING_STATUS_SENDING);
+		$count = 0;
+		$mailing->setAttribute('progress', $count);
```
matthijskooijman, line 1066: This should use a constant. Same for `status` btw, and some other string constants in this function, but those would be a separate comit (would be good if you could fix them up, though).

matthijskooijman (commented):
I left some nitpick comments, overall it looks great. I also tested, worked like a charm here :-)

Regarding commit structure:
 - I would put commit "Add css changes for mailing progress ui" directly after "Display progress for mailinglist send" since it is tightly coupled.
 - The commit message for "Display progress for mailinglist send" still has some old points included and is an enumeration of points which aren't terribly clear on their own. Could you update that to a self-standing commit message (good to name the subparts of the commit, but maybe with a few more words).

system/core/various.php
```diff
@@ -97,7 +97,7 @@ function asterisk($condition) {
 	*/
 	function makeButton($label, $action, $id = '', $class = '') {
 		$class = 'button' . ($class ? ' ' . $class : '');
-		return '<input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'.$action.'" />';
+		return '<input type="button" class="'.$class.'" '.($id?' id="'.$id.'"':'').'value="'.$label.'" onclick="'.htmlspecialchars($action).'" />';
```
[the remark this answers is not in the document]
RoukePouw: Done :)

system/datatypes/mailinglist.php
RoukePouw: Done :)

system/datatypes/mailinglist.php
RoukePouw: Done :)

system/datatypes/mailinglist.php
RoukePouw: Done :)

matthijskooijman:
I made a few more changes:
 - Added "mailinglist: " prefix to commit message
 - Converted a few more `'progress'` to `self::FIELD_NAME_PROGRESS`
 - Converted a few other strings to constants too (as a separate commit) and moved one conversion of `'status'` to this new commit.
 - Fixed one incorrect type annotation I noticed (was already broken in master).
 - Rebased on top of master.

 With that, I'll merge this.

matthijskooijman (approved):

Issue #359, opened 2020-12-07
Refactor to use multiple <form> tags


matthijskooijman:
Currently, there is only one `<form>` tag in the entire page that is shared by all forms, which is not always ideal (i.e. in peer_reviewed_article, this requires adding a suffix to field names since *all* comment forms are submitted at the same time).

In #135, some changes were made to prepare for having multiple form tags (i.e. form submissions now explicitly reference their containing form and an explicit "default form" is tracked on HyphaHtmlDocument). This should be completed at some point.

A related improvement would be to turn submit buttons into normal submit buttons, rather than doing all form submits through the javascript `hypha()` function. This makes things simpler, and allows the browser to apply any validation (i.e. from `required` or `type=email` attributes on input fields, which are currently ignored, presumably because javascript forces form submission).

PR #360, opened 2020-12-07
Refactor form validation

matthijskooijman:
The main point of this PR is the last commit, which guards against insertion of `<script>` tags in HTML fields. The other commits refactor the form input validation so there is now a single `validate()` method that must always be called, where this check can be done automatically on all wymeditor fields (previously, validation always required calling `validateXXXField()` methods explicitly). These refactors were something I had been considering for a while, since they make the form generation again a little bit more declarative and automatic.


system/core/document.php
```diff
@@ -568,11 +569,52 @@ public function dataFor($name, $default = null) {
...
+		 *  - Fields with the `required` attribute set.
+		 *  - Fields with the `type=email` set.
+		 *  - Editor fields for safe HTML
+		 */
+		function validate() {
+			foreach ($this->fields as $name => $elems) {
+				foreach ($elems as $elem) {
+					if ($elem->hasAttribute('required'))
+						$this->validateRequiredField($name);
+					if ($elem->tagName == 'input' && $elem->getAttribute('type') == 'email')
```
RoukePouw, line 598: If an email field is empty I think you would like the message 'required field missing', not 'incorrect email address' which seems to happen now that the first error gets overwritten by the second? Perhaps shift the order? Or make the errors per name into an array instead?

system/languages/en.php
```diff
@@ -184,6 +184,8 @@
 		"edit-day" => "edit day",
 		"invalid-money-amount" => "Not a valid amount of money",
 		"required-field-missing" => "Required field missing",
+		"email-field-invalid" => "Invalid email address supplied",
+		"html-field-contains-script" => "HTML cannot contain <script> tag. Your computer or browser might be infected by a virus",
```
RoukePouw, line 188: I find this message slightly confusing..
If I as a user would copy paste some html I wanted to use and I get a notification that my computer might be infected by a virus. I would get worried, perhaps for the wrong reasons. 
Something along these lines perhaps? `"HTML <script> tags are not allowed. They are considered unsafe as they can be used for hacks or virusses."`

system/core/document.php
```diff
@@ -581,13 +623,50 @@ function validateRequiredField($name) {
...
+		 * Normally, you should use the `editor` field to make
+		 * validate() apply this check automatically instead of
+		 * calling this directly.
+		 */
+		function validateHtmlField($name) {
+			$value = $this->dataFor($name);
+			if (!$value)
+				return true;
+
+			$dom = $this->root->document()->createElement('dummy');
```
RoukePouw, line 660: Does this need to be cleaned up afterwards?

system/languages/en.php
RoukePouw: Ah, just read this comment in the code above, this explains the reasoning behind the stringent message. But I think the message could still be made more clear.

```
 (and these have been seen to be
 			// automatically inserted into requests by
 			// viruses...).
```

system/core/document.php
RoukePouw: Ah, no because the an empty email field is considered valid, the error won't be overwritten.

matthijskooijman: Exactly.

system/core/document.php
```diff
@@ -568,11 +569,52 @@ public function dataFor($name, $default = null) {
...
-			if (!array_key_exists($name, $this->fields)) {
-				$this->errors[$name] = __('field-not-found');
-				return false;
+		/**
+		 * Is this form valid? Returns true when:
+		 *  - validate() was called
+		 *  - all of the fields for which validateXXXField
+		 *    methods were called were also valid.
+		 */
+		function isValid() {
```
RoukePouw, line 578: Executing the form validate() here, if not yet done, would reduce some code duplication. It would loose it's const-ness but it would perform more as the name suggests.

system/languages/en.php
matthijskooijman: Yeah, I didn't spend too much effort on the message, since it should not normally be shown. Do you have a specific suggestion for a better message?

system/core/document.php
matthijskooijman: I don't think so, since it is created but not added into the document tree (so I'm assuming PHP will garbage collect the DomNode itself).

system/core/document.php
matthijskooijman: Interesting idea, though I'm not entirely sure. Always running validate would mean a form can be validated more than once (which is a waste of efffort), but running it only if not ran before means that re-validating after a change in the form is not possible (we don't really use or support this now, though).

But maybe running it when not ran before is a good idea. If we ever need to re-validate, you could just call `validate()` explicitly then. I'll see how to best implement this, since it might nicest to fold this change back into the git history (but maybe that's too much work and I'll just add a new commit).

system/languages/en.php
RoukePouw: Something like this? 
"HTML <script> tags are not allowed. They are considered unsafe as they can be used for hacks or viruses. If you not add them yourself your computer could even be infected."

system/core/document.php
laurensmartina: If it would self validate, then, when a value gets updated, the is_validated flag could be set to false.

laurensmartina (approved):
Very cool improvement!

system/languages/en.php
matthijskooijman: Thanks, I've used your suggestion.

system/core/document.php
matthijskooijman: Ok, I implemented automatic validation like suggested. Re-validating / clearing `is_validated` on changes is not yet implemented, since there is no meaningful way to change values right now I believe, so this is left for later.

matthijskooijman:
Rebased on top of master and made some small changes based on review, going to review next.

matthijskooijman:
Hm, turns out this has an unintended side effect: Because everything lives in one big form, validation errors in e.g. the comment form let the browser prevent submitting the login form.

For now, I've fixed this by reverting this PR on www.destadsbron.nl, but the proper fix for this is probably to separate each form into its own `<form>` tag, which is a refactor that we were planning to do anyway (which is #359).

PR #364, opened 2020-12-23
Improve comment links

matthijskooijman:
This adresses some problems with the comment links feature merged in #356, as discussed in https://github.com/PlanBCode/hypha/pull/356#issuecomment-743437016 onwards.

This PR contains the commits previously pushed by @RoukePouw, but I:
 - Split his first commit into two
 - Added a fix for php-dom-wrapper's `create` method eating up text not wrapped in a span or so. I've suggested this same fix upstream, and I'd like to hear their response before merging this here (ideally, we could just backport a fix with an upstream commit reference as we've done before). See https://github.com/scotteh/php-dom-wrapper/issues/15#issuecomment-750145412
 - Limited the regex to require either http://, https:// or a `www` prefix, to prevent matching words with a dot halfway.
 - Added `_` to the allowed characters in urls
 - Removed the non-greedy `*?` matching and removed the `+` from the trailing non-dot, since that would prevent a domain with three components (e.g. https://www.dewar.nl) from being matched.
 - Removed one reference to the now-removed `$suffix` variable.
 - Rebased on top of master.

@RoukePouw, how does this look? If this looks ok to you and the php-dom-wrapper maintainer applies this fix upstream, I think this addresses everything?

### DIFF
--- a/system/datatypes/peer_reviewed_article.php
+++ b/system/datatypes/peer_reviewed_article.php
@@ -1452,8 +1452,7 @@ EOF;
 		$commentHtml = htmlspecialchars($comment->getText());
 		$tryToParseLinks = preg_replace_callback(self::URL_REGEX, function($matches) { // replace urls with links
 			$url = $matches[0];
-			// encapsulated in a span to prevent trimming by html->create
-			return '<span><a href="' . $url . '" target="_blank">' . $url . '</a></span>';
+			return '<a href="' . $url . '" target="_blank">' . $url . '</a>';
 		}, $commentHtml);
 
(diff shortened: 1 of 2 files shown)


RoukePouw:
Looks great! :)

matthijskooijman:
I've rebased this on master, and squashed the fixups. I want to go ahead and merge this even though phpdomwrapper's maintainer has not responded yet, we can always update our fix when they do.

@RoukePouw, I also noticed that in one of your commits, you removed the `u` modifier on the regex, which enables UTF-8 processing. Was that intentional? It looks like having that modifier is correct, so I modified your commits to leave it now.

Issue #361, opened 2020-12-23
Shorten long comments and add 'Read more...'


dianawi:
Some of the comments are very long, causing readers to have to scroll to get to the next. 
For comments longer than 100 words, cut the text after 70 words and add 'read more..,' link that shows the whole text.

PR #366, opened 2020-12-24
Shorten long comments and add Read more...

RoukePouw:
https://github.com/PlanBCode/hypha/issues/361

Longer comments get cut into a (html-less) stub (to prevent broken html) and appended by a 'read more' button.

### DIFF
--- a/system/datatypes/peer_reviewed_article.php
+++ b/system/datatypes/peer_reviewed_article.php
@@ -130,7 +130,8 @@ class peer_reviewed_article extends HyphaDatatypePage {
 
 	const URL_REGEX='/(https?:\/\/|www.)[\w\/-]+\.[\w\/-?=%.&#_]*[\w\/-?=%&#_]/iu';
 	// [protocol://|www.] [word] . [suffix /path ?query&parameter=value#bookmark] [last char must not be a dot]
-	const MAX_COMMENT_LENGTH = 100;
+	const MAX_COMMENT_LENGTH = 100 * 6; // Roughly hundred words
+	const MIN_COMMENT_STUB_LENGTH = self::MAX_COMMENT_LENGTH  * 0.7;
 
 	public static function getDatatypeName() {
 		return __('datatype.name.peer_reviewed_article');
@@ -1438,6 +1439,19 @@ EOF;
 		return $approvesDomElement;
 	}
 
+	protected function processCommentText($commentText){
+		$commentHtml = htmlspecialchars($commentText);
+		$tryToParseLinks = preg_replace_callback(self::URL_REGEX, function($matches) { // replace urls with links
+			$url = $matches[0];
+			return '<a href="' . $url . '" target="_blank">' . $url . '</a>';
+		}, $commentHtml);
+
+		if(!is_null($tryToParseLinks)) $commentHtml = $tryToParseLinks; // In case regex fails
+
+		$commentHtml = nl2br($commentHtml);
+		return $commentHtml;
+	}
+
 	/**
 	 * Render a single comment.
 	 *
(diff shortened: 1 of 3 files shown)


system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
 			return '<span><a href="' . $url . '" target="_blank">' . $url . '</a>' . $suffix.'</span>';
 		}, $commentHtml);
 		$commentHtml = nl2br($commentHtml);
+		if(strlen($commentHtml) > self::MAX_COMMENT_LENGTH){
+			$id = rand();
+			$stub = substr(strip_tags($commentHtml),0, self::MAX_COMMENT_LENGTH*0.7);
```
matthijskooijman, line 1468: This probably should not hardcode 0.7, but just use a second constant?

system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
...
+			$commentHtml = '<div style="display:none;" id="commentFull'.$id.'">'.$commentHtml.'</div>
+			<span id="commentStub'.$id.'">'.$stub.'...</span>
```
matthijskooijman, line 1470: Why use `$commentHtml` here and strip_tags that? That effectively just cancels the link-detection and nl2br?

Also even without HTML tags, there's currently a chance of cutting halfway an entity (i.e. `&amp;`) which could break the HTML (or at least result in an unsightly `&amp` or so being shown).

Wouldn't it be better to cut the original unprocessed comment and then apply all processing (htmlspecialchars, link detection, nl2br) all on the stub? AFAICS that should result in consistent and safe output, that has newlines in all the places you'd expect. The only caveat I can think of is that links might be cut halfway, which could be fixed by skipping the linkifying for a match at the very end of the comment, or by cutting on spaces rather than halfway words too.

Of course this does require that you remember that a comment was cut so you can add the "read more" link *after* all the processing, rather than before.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
 			return '<span><a href="' . $url . '" target="_blank">' . $url . '</a>' . $suffix.'</span>';
 		}, $commentHtml);
 		$commentHtml = nl2br($commentHtml);
+		if(strlen($commentHtml) > self::MAX_COMMENT_LENGTH){
```
matthijskooijman, line 1466: This limits to 100chars, while @dianawi suggested cutting on 100 *words*. Also, cutting on whitespace only probably results in a nicer-to-read result as well, and might solve an issue with cutting links halfway (see below).

system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
...
+			$id = rand();
+			$stub = substr(strip_tags($commentHtml),0, self::MAX_COMMENT_LENGTH*0.7);
+			$commentHtml = '<div style="display:none;" id="commentFull'.$id.'">'.$commentHtml.'</div>
+			<span id="commentStub'.$id.'">'.$stub.'...</span>
+			<br/><a id="commentToggle'.$id.'" onclick="
+				document.getElementById(\'commentStub'.$id.'\').style.display=\'none\';
+				document.getElementById(\'commentFull'.$id.'\').style.display=\'block\';
+				document.getElementById(\'commentToggle'.$id.'\').style.display=\'none\';
```
matthijskooijman, line 1474: It might be nicer to add some classes and/or `data` attributes here and then put the actual js in a more global spot. OTOH, this is a pragmatic and self-contained approach, of course.

system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
 			return '<span><a href="' . $url . '" target="_blank">' . $url . '</a>' . $suffix.'</span>';
 		}, $commentHtml);
 		$commentHtml = nl2br($commentHtml);
+		if(strlen($commentHtml) > self::MAX_COMMENT_LENGTH){
+			$id = rand();
```
matthijskooijman, line 1467: This seems a bit fragile. I wonder if the javascript couldn't just find the right bits based on searching nearby the link? I.e. search for a certain class inside the parent element? IIRC we have jquery inluded everywhere, so that would probably make this easier?

system/datatypes/peer_reviewed_article.php
```diff
@@ -1462,6 +1463,17 @@ protected function createCommentDomElement(HyphaDomElement $comment, $review = f
...
+			$stub = substr(strip_tags($commentHtml),0, self::MAX_COMMENT_LENGTH*0.7);
+			$commentHtml = '<div style="display:none;" id="commentFull'.$id.'">'.$commentHtml.'</div>
+			<span id="commentStub'.$id.'">'.$stub.'...</span>
+			<br/><a id="commentToggle'.$id.'" onclick="
+				document.getElementById(\'commentStub'.$id.'\').style.display=\'none\';
+				document.getElementById(\'commentFull'.$id.'\').style.display=\'block\';
+				document.getElementById(\'commentToggle'.$id.'\').style.display=\'none\';
+			">' . __('art-read-more') . '...</a>'; //TODO translation
```
matthijskooijman, line 1475: This TODO is already implemented, right?

matthijskooijman (changes requested):
Looks like a good start. I did leave "a few" comments inline :-p

system/datatypes/peer_reviewed_article.php
RoukePouw: Done, thanks :)

system/datatypes/peer_reviewed_article.php
RoukePouw: Changed the process according to your feedback :)

system/datatypes/peer_reviewed_article.php
RoukePouw: Thanks! Switched to 100 * 6 chars which, on average, comes down to roughly 100 words.

system/datatypes/peer_reviewed_article.php
RoukePouw: Correct, removed :)

system/datatypes/peer_reviewed_article.php
RoukePouw: Done, vanilla js could do the trick as well :)

system/datatypes/peer_reviewed_article.php
matthijskooijman: Sounds pragmatic :-)

system/datatypes/peer_reviewed_article.php
matthijskooijman: Looks good!

system/datatypes/peer_reviewed_article.php
matthijskooijman: Nice!

system/datatypes/peer_reviewed_article.php
matthijskooijman: Looks great, thanks!

system/datatypes/peer_reviewed_article.php
matthijskooijman: Ok, let's leave it like this.

matthijskooijman:
Ok, all comments have been adressed. I've additionally made a few more small changes:
 - Rebase on master
 - Fix some style issues
 - Add CSS classes
 - Use "" instead of '' so \n is actually interpreted correctly.
 - Remove <br/>, this can be done in theme CSS if required

With that, I think this is ready to merge.

PR #369, opened 2020-12-30
peer_reviewed_article: Fix use of - in links in comments

matthijskooijman:
The - was already included in the URL regex used to detect links in
comments, but because it was not at the start of the `[...]` regex
element, it was probably interpreted as a range dash rather than a
literal dash. Putting it at the front fixes this and allows dashes in
links as intended.


matthijskooijman:
@RoukePouw, I found one more small problem with the link detection, fixed by this PR. Since it is quite trivial, I'm going to merge this right away.

PR #371, opened 2021-01-20
wymeditor: Support audio and video in embed plugin

matthijskooijman:
This is partly based on https://gist.github.com/winhamwr/2956136 but
with audio added and allowing all attributes documented by w3schools for
these tags. Also make video and audio a block tag, so source can be
nested inside.

### DIFF
--- a/system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
+++ b/system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
@@ -58,6 +58,53 @@
             ]
         };
 
+        WYMeditor.XhtmlValidator._tags.video = {
+            "attributes":[
+                "width",
+                "height",
+                "autoplay",
+                "controls",
+                "loop",
+                "muted",
+                "poster",
+                "preload",
+                "src"
+            ]
+        };
+
+        WYMeditor.XhtmlValidator._tags.audio = {
+            "attributes":[
+                "autoplay",
+                "controls",
+                "loop",
+                "muted",
+                "preload",
+                "src",
+                "crossorigin",
+                "currentTime",
+                "disableRemotePlayback"
+            ]
+        };
+
+        WYMeditor.XhtmlValidator._tags.source = {
+            "attributes":[
+                "src",
+                "srcset",
+                "type",
+                "media",
+                "sizes",
+                "autoPictureInPicture",
+                "buffered",
+                "controlslist",
+                "crossorigin",
+                "currentTime",
+                "disablePictureInPicture",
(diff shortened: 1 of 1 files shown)


system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
```diff
@@ -58,6 +58,40 @@
...
+                "controls",
+                "loop",
+                "muted",
+                "preload",
+                "src"
+            ]
+        };
+
+        WYMeditor.XhtmlValidator._tags.source = {
+            "attributes":[
```
RoukePouw, line 87: there seems to be a srcset attribute as well
source: (:-p) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
```diff
@@ -58,6 +58,40 @@
...
+                "controls",
+                "loop",
+                "muted",
+                "poster",
+                "preload",
+                "src"
+            ]
+        };
+
+        WYMeditor.XhtmlValidator._tags.audio = {
```
RoukePouw, line 75: more attributes: crossorigin, currentTime, duration, disableRemotePlayback (Experimental)  
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio

system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
```diff
@@ -58,6 +58,40 @@
             ]
         };
 
+        WYMeditor.XhtmlValidator._tags.video = {
+            "attributes":[
```
RoukePouw, line 62: more attributes: autoPictureInPicture, buffered, controlslist, crossorigin, currentTime, disablePictureInPicture, disableRemotePlayback, duration,intrinsicsize , playsinline
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

system/wymeditor/plugins/embed/jquery.wymeditor.embed.js
matthijskooijman: I saw that, but according to w3schools that was only valid when using `<picture>`, which we don't support anyway. But, reconsidering, might be better to just add this now and make it easier to add `<picture>` later.

matthijskooijman:
I added the missing attributes (except for `duration`, which is marked as readonly which I think means it can be read from javascript but not set in HTML).

Will merge next.

PR #372, opened 2021-01-27
Set byte range headers for audio and video files

RoukePouw:


system/core/database.php
```diff
@@ -560,6 +560,44 @@ function serveFile($filename, $root) {
 			die("Invalid filename: $filename");
 		}
 		header('Content-Type: ' . getMimeType($filename));
+	  $filesize = filesize($filename);
+		$shortlen =$filesize-1;
+
+		if(getMimeType($filename)==='audio/mpeg'){
```
matthijskooijman: No need to limit on mimetype, right? If the browser asks for a part of any file, it can just get it?

system/core/database.php
```diff
@@ -560,6 +560,44 @@ function serveFile($filename, $root) {
...
+						$start = intval($x[0]);
+						$end = intval($x[1]);
+						http_response_code(206);
+						$bytesize=$end-$start;
+						header( 'Content-Length: ' . (string)($bytesize));
+						header( 'Content-Range: bytes '.$start.'-'.$end.'/'.$filesize);
+						header("Pragma: public");
+				 		header("Expires: 0");
+					  header( 'Accept-Ranges: bytes');
+						header('X-Pad: avoid browser bug');
```
matthijskooijman: If you have it, please add a reference for this workaround.

system/core/database.php
```diff
@@ -560,6 +560,44 @@ function serveFile($filename, $root) {
...
+						$x = explode('-',$range); // "start-end" -> ["start","end"]
+						$start = intval($x[0]);
+						$end = intval($x[1]);
+						http_response_code(206);
+						$bytesize=$end-$start;
+						header( 'Content-Length: ' . (string)($bytesize));
+						header( 'Content-Range: bytes '.$start.'-'.$end.'/'.$filesize);
+						header("Pragma: public");
+				 		header("Expires: 0");
+					  header( 'Accept-Ranges: bytes');
```
matthijskooijman: This should probably also be added when the full file is requested?

system/core/database.php
```diff
@@ -560,6 +560,44 @@ function serveFile($filename, $root) {
...
+						header('X-Pad: avoid browser bug');
+					  header('Cache-Control: no-cache');
+						header('Content-Disposition: inline; filename="' . 'test.mp3' . '"');
+
+						$handle = fopen($filename, "r");
+						$contents = fread($handle, $end);
+						fclose($handle);
+						echo substr($contents,$start);
```
matthijskooijman: This is really non-ideal, better to not read everything (see e.g. https://gist.github.com/codler/3906826 or https://www.php.net/manual/en/function.readfile.php#86244).

Even better would be to just tell PHP to handle all the reading/writing (i.e. like `readfile` but for a part of the file), so that not all of the data has to pass through PHP variables, but I couldn't quickly find any way to do that (there is `passthru()`, but that always reads until EOF).

matthijskooijman (commented):
I left some quick thoughts inline, I realize that there's more to cleanup.

Overall, I would think a code structure something like this would be good:
```
$filesize = ...
if (has_range_header) {
  $start = ...
  $end = ...
} else {
  $start = 0;
  $end = 0;
}

// Send all kinds of common headers
if ($start == 0 && $end == $filesize - 1) {
  readfile(...);
} else {
  // More complex reading code
}
```
This allows not duplicating the common headers and keeps a single exit point from this code, which is probably better.

system/core/database.php
RoukePouw: Check

system/core/database.php
RoukePouw: Agreed :-)

system/core/database.php
RoukePouw: Agreed, haven't found a way to do that yet, will do some searching

RoukePouw: fseek https://stackoverflow.com/a/16741352  but that still needs to pass through a variable

matthijskooijman: Yeah, I suspect that's as good as it gets. Have a look at my two links, which do a block-based approach which is *probably* more efficient that loading *all* data into memory at once with `fgets` like in your link.

RoukePouw: I now have these two working:
```
fseek($handle,$start,0);
print fread($handle,$bytesize);
```
or

```			
$cur=$begin;
fseek($handle,$begin,0);
while(!feof($handle)&&$cur<$end&&(connection_status()==0)){ 
  print fread($handle,min(1024*16,$end-$cur));
  $cur+=1024*16;
}
```

Do I understand correctly that you think the second one would be better?

matthijskooijman: That's correct.

Though the the direct `print(fread(...))` combination *could* allow PHP to optimize this and never store the result in PHP variable (in which case the first approach would be fine), I'm afraid that it won't and ends up reading everything into memory first.

system/core/database.php
```diff
@@ -559,8 +559,48 @@ function serveFile($filename, $root) {
...
+		header('Content-Range: bytes '.$start.'-'.$end.'/'.$filesize);
+		header('Accept-Ranges: bytes');
+		header("Pragma: public"); // TODO not sure if really needed
+		header('X-Pad: avoid browser bug'); // TODO not sure if really needed
+		header('Cache-Control: no-cache'); // TODO not sure if really needed
+		header('Content-Disposition: inline; filename="' . $filename . '"'); //TODO only basename?
+
+		if($start === 0 && $end === $filesize){
+			readfile($filename);
+		}else{
```
matthijskooijman: This should be `if (...) {` and `} else {` (i.e. more spaces).

system/core/database.php
```diff
@@ -559,8 +559,48 @@ function serveFile($filename, $root) {
 			http_response_code(404);
 			die("Invalid filename: $filename");
 		}
+	  $filesize = filesize($filename);
```
matthijskooijman: More mixed spaces and tabs here and elsewhere.

system/core/database.php
```diff
@@ -559,8 +559,48 @@ function serveFile($filename, $root) {
...
+		if($start === 0 && $end === $filesize){
+			readfile($filename);
+		}else{
+			http_response_code(206); // Partial content
+			$handle = fopen($filename, "r");
+			/*Used a chunk approach for printing instead of
+			fseek($handle,$start,0);
+			print fread($handle,$bytesize);*/
+
+			$cur=$begin;
```
matthijskooijman: This says `$begin`, but it should be `$start`. Did you test the last push?

matthijskooijman (commented):
I left some comments inline, but also fixed them myself in a new commit.

I also wondered about the `Content-disposition` header, AFAICT if you make a link to files now, that forces the file to be displayed in the browser (unlike offering it for download, as you might want in some cases). Or is `inline` maybe already the default, displaying files in the browser or downloading when the browser cannot display the file? In any case, IIRC there's no way to set the filename without *also* setting the disposition, so there might not be an alternative.

As discussed on Element, I'm not yet sure that this change (even though it's good on its own) is actually fixing the right problem.

matthijskooijman:
header("Pragma: public"); // TODO not sure if really needed

This is apparently some implementation-specific header related to caching in some implementations. `Pragme: no-cache` is standard, but other values are implementation-defined.

https://stackoverflow.com/questions/1920781/what-does-the-http-header-pragma-public-mean

                  header('X-Pad: avoid browser bug'); // TODO not sure if really needed

This is apparently a workaround for a bug in Netscape, fixed 24 years ago, but the header still lingered in Apache for quite a while.
https://stackoverflow.com/questions/8711584/x-pad-avoid-browser-bug-header-added-by-apache
https://github.com/mnot/redbot/issues/229

                  header('Cache-Control: no-cache'); // TODO not sure if really needed

This prevents using cached resources without first validating them. I'd think this is not really needed, since the browser is free to validate partial responses, just like complete responses. However, it seems that Apache (or PHP) actually adds `Cache-Control: no-store` (which disallows *all* caching) on all responses by default, so I guess this header is intended to actually *allow* some caching. However, since we support no way of validation (`If-modified-Since` or `Etag` headers), `no-cache` effectively prevents all caching, which doesn't seem particularly useful. So just setting this to `public` with a `max-age` of 24h is probably better.

So, I've removed `X-pad` and updated the cache heades. I've also removed the `Content-Disposition` header, since that primarily sets the filename, but the browser can already just derive the filename from the URL. And I added a TODO about Apache's `X-Sendfile` header, to be looked at later.

With this, I think the code is ready to be tested, and if it works as expected, merged (after squashing).

system/core/database.php
```diff
@@ -559,8 +559,52 @@ function serveFile($filename, $root) {
...
 			die("Invalid filename: $filename");
 		}
+		// TODO: Use X-Sendfile header to let Apache serve the file?
+		$filesize = filesize($filename);
+		$headers = getallheaders();
+		$matches = [];
+		// Only supports fully-qualified bytes=start-end ranges,
+		// not `bytes=-suffixlength` or `bytes=start-` or
+		// multi-part ranges.
+		if (array_key_exists('Range', $headers) && preg_match('/bytes=(\d+)-(\d+)$/', $headers['Range'], $matches)) {
```
RoukePouw, line 569: should it give an error if a multi part is requested but could not be returned?

matthijskooijman: I'm not sure what error would be appropriate here. If we just return the entire file, the client could see the absence of the range headers and conclude that it wasn't supported, I hope...

matthijskooijman: Seems ignoring the header is always allowed: https://stackoverflow.com/a/6204397/740048

matthijskooijman:
I've squashed the commits and rebased on top of master, will merge next.

